PR #101, opened 2022-12-25
feat: automatic onion services

juanAngel:
## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM node:18-alpine3.16 as build
 
 WORKDIR /build
-RUN apk add --no-cache --update git
 
 COPY ["package.json", "package-lock.json", "./"]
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -26,11 +27,13 @@ ENV DB_USER=nostr-ts-relay
 ENV DB_PASSWORD=nostr-ts-relay
 
 WORKDIR /app
+RUN apk add --no-cache --update git
```
Cameri, line 30: why is `git` needed?

juanAngel: At first it was, but now it's not.

juanAngel:
Do you have any idea why the commit linter fails?
Since changes cannot be uploaded without complying with this. It's not something that should fail

Issue #75, opened 2022-12-16
Support NIP-40 Expiration Timestamps


Cameri:
https://github.com/nostr-protocol/nips/pull/87

Cameri:
Completed in #148

PR #148, opened 2023-01-28
feat: NIP-40

antonleviathan:
This is a PR to implement [NIP-40](https://github.com/nostr-protocol/nips/blob/master/40.md)

## Description
Implementing a method which skips events whose expiration time is equal or greater than the current time 

## Related Issue
https://github.com/Cameri/nostream/issues/75

## Motivation and Context
This is a part of NIPs

## How Has This Been Tested?
Tests are yet to be written (WIP)

## Types of changes

## Checklist:


antonleviathan:
Hey @Cameri here is the WIP for NIP-40! I wanted to get your feedback on the direction I've taken so far and to get your help regarding where in the event respository to skip the expired event!

src/handlers/event-strategies/default-event-strategy.ts
```diff
@@ -14,6 +15,24 @@ export class DefaultEventStrategy implements IEventStrategy<Event, Promise<void>
     private readonly eventRepository: IEventRepository,
   ) { }
 
+  protected async isEventExpired(event: Event): Promise<Event | ExpiredEvent> {
+    const [, rawExpirationTime] = event.tags.find((tag) => tag.length === 2 && tag[0] === EventTags.Expiration)
```
Cameri, line 19: should probably check that the tag length is >= 2, the rest of the elements are safe to ignore but the event would still be valid

Cameri: would add `Number.isSafeInteger(Number(tag[1]))`

PR #202, opened 2023-02-13
feat: add LNURL processor

im-adithya:
## Description

Adds an lnurl processor for payments. We used the LNURL verify spec: https://github.com/lnurl/luds/issues/182

This is made in such a way that any provider can implement, all they have to do is return a verify link in their lnurlp callback. 

For example:
https://getalby.com/lnurlp/adithya/callback?amount=10000

```
{
    "status": "OK",
    "verify": "https://getalby.com/lnurlp/adithya/verify/HLzzSn7vzrAwgrsmTJ9LjQSH",  // <--
    "routes": [],
    "pr": <bolt11 invoice>
}
```

And a call to the verify URL returns the status:
```
{
    "status": "OK",
    "settled": false, // <--
    "preimage": null,
    "pr": <bolt11 invoice>
}
```

To support this, we added a `verify_url` field in the database.

## Motivation and Context

Mitigates the need of adding processors for each and every provider separately.

## How Has This Been Tested?

Tested it locally.

Screenshot 2023-02-17 at 2 31 48 PM

## Types of changes

## Checklist:


im-adithya:
@Cameri can you please review this :D

Cameri:
> @Cameri can you please review this :D

Thanks for contributing. It's on my todo list.

Cameri:
@im-adithya could you please rebase and fix the conflicts?

im-adithya:
Done!

Cameri:
@im-adithya do you have a relay instance running with this where you and I can test this code?

im-adithya:
Sorry, I don't :(
I just tested it locally

src/payments-processors/lnbits-payment-processor.ts
```diff
@@ -104,13 +104,13 @@ export class LNbitsPaymentsProcesor implements IPaymentsProcessor {
 
       debug('response: %o', response.data)
 
-      const invoiceResponse = await this.httpClient.get(`/api/v1/payments/${encodeURIComponent(response.data.payment_hash)}`, {
+      const invoiceResult = await this.httpClient.get(`/api/v1/payments/${encodeURIComponent(response.data.payment_hash)}`, {
```
Cameri, line 107: can we keep this variable named `invoiceResponse`?

src/constants/base.ts
```diff
@@ -40,6 +40,7 @@ export enum EventTags {
 }
 
 export enum PaymentsProcessors {
+  LNURL = 'lnurl',
```
Cameri, line 43: this has been added to settings, but not to `resources/default-settings.yaml`
please add default settings for this payment processor

src/controllers/invoices/post-invoice-controller.ts
```diff
@@ -165,7 +165,7 @@ export class PostInvoiceController implements IController {
       relay_url: relayUrl,
       pubkey,
       relay_pubkey: relayPubkey,
-      expires_at: invoice.expiresAt?.toISOString(),
+      expires_at: invoice.expiresAt?.toISOString() || '',
```
Cameri, line 168:
```suggestion
      expires_at: invoice.expiresAt?.toISOString() ?? '',
```

src/payments-processors/lnbits-payment-processor.ts
```diff
@@ -45,29 +45,29 @@ export class LNbitsPaymentsProcesor implements IPaymentsProcessor {
...
-    debug('get invoice: %s', invoiceId)
+  public async getInvoice(invoice: Invoice): Promise<GetInvoiceResponse> {
+    debug('get invoice: %s', invoice.id)
     try {
-      const response = await this.httpClient.get(`/api/v1/payments/${invoiceId}`, {
+      const response = await this.httpClient.get(`/api/v1/payments/${invoice.id}`, {
         maxRedirects: 1,
       })
-      const invoice = new LNbitsInvoice()
+      const invoiceResult = new LNbitsInvoice()
```
Cameri, line 54: do we need to rename this? `invoice` seemed fine

src/payments-processors/lnurl-payments-processor.ts
```diff
@@ -0,0 +1,69 @@
+import { AxiosInstance } from 'axios'
+import { Factory } from '../@types/base'
+
+import { CreateInvoiceRequest, GetInvoiceResponse, IPaymentsProcessor } from '../@types/clients'
+import { Invoice, InvoiceStatus, InvoiceUnit } from '../@types/invoice'
+import { createLogger } from '../factories/logger-factory'
+import { randomUUID } from 'crypto'
+import { Settings } from '../@types/settings'
+
+const debug = createLogger('alby-payments-processor')
```
Cameri, line 10: `alby`?

src/constants/base.ts
im-adithya: Done!

src/payments-processors/lnurl-payments-processor.ts
```diff
@@ -0,0 +1,69 @@
...
+      amount: amountMsats,
+      description,
+      requestId,
+    } = request
+
+    try {
+      const response = await this.httpClient.get(`${this.settings().paymentsProcessors?.lnurl?.invoiceURL}/callback?amount=${amountMsats}&comment=${requestId}`)
+
+      const result = {
+        id: randomUUID(),
```
Cameri, line 47: although this works, I'm concerned we won't be able to look up this invoice on your LNURL provider. Should we instead use UUID v5 which uses namespaces and can be used to generate deterministic UUIDs? That way we can pass the invoice description and always get the same UUID back.

src/payments-processors/lnurl-payments-processor.ts
```diff
@@ -0,0 +1,69 @@
...
+  public async createInvoice(request: CreateInvoiceRequest): Promise<any> {
+    debug('create invoice: %o', request)
+    const {
+      amount: amountMsats,
+      description,
+      requestId,
+    } = request
+
+    try {
+      const response = await this.httpClient.get(`${this.settings().paymentsProcessors?.lnurl?.invoiceURL}/callback?amount=${amountMsats}&comment=${requestId}`)
```
Cameri, line 44: should `comment` be `request.description`?

src/payments-processors/lnurl-payments-processor.ts
im-adithya: We were first trying to bring support for [Alby](https://getalby.com/) but then realized we can generalize this for everyone implementing LNURL-verify, changing it to lnurl!

src/payments-processors/lnurl-payments-processor.ts
```diff
@@ -0,0 +1,69 @@
...
+
+      const result = {
+        id: randomUUID(),
+        pubkey: requestId,
+        bolt11: response.data.pr,
+        amountRequested: amountMsats,
+        description,
+        unit: InvoiceUnit['MSATS'],
+        status: InvoiceStatus['PENDING'],
```
Cameri, line 53:
```suggestion
        unit: InvoiceUnit.MSATS,
        status: InvoiceStatus.PENDING,
```

src/repositories/invoice-repository.ts
```diff
@@ -103,6 +103,11 @@ export class InvoiceRepository implements IInvoiceRepository {
         always(undefined),
         prop('createdAt'),
       ),
+      verify_url: ifElse(
+        propSatisfies(isNil, 'verifyURL'),
+        always(undefined),
+        prop('verifyURL'),
+      ),
```
Cameri, line 110:
```suggestion
      verify_url: prop('verifyURL'),
```
essentially the same :)

src/services/payments-service.ts
```diff
@@ -111,17 +113,11 @@ export class PaymentsService implements IPaymentsService {
...
     try {
+      const fullInvoice = await this.invoiceRepository.findById(invoice.id)
       await this.invoiceRepository.upsert({
-        id: invoice.id,
-        pubkey: invoice.pubkey,
-        bolt11: invoice.bolt11,
-        amountRequested: invoice.amountRequested,
-        description: invoice.description,
-        unit: invoice.unit,
+        ...fullInvoice,
```
Cameri, line 118: let's revert this change. `upsert` already updates some invoice properties if one already exists in the database.

src/payments-processors/lnurl-payments-processor.ts
im-adithya: We didn't have a uuid string coming from the response: https://getalby.com/lnurlp/adithya/callback?amount=10000
So we thought of using a random UUID.

> instead use UUID v5 which uses namespaces and can be used to generate deterministic UUIDs?

The problem is that we don't return the description (and because of that I changed the getInvoice as well to take the whole invoice instead of just invoice.id)

src/payments-processors/lnurl-payments-processor.ts
Cameri: lovely!

src/payments-processors/lnurl-payments-processor.ts
Cameri: well, this could be a future improvement, not a blocker. But we definitely want to give relay operators the ability to go to their LNURL provider to lookup invoices. If they can at least see descriptions which should include the kind of fee and pubkey then it's all good for now.

src/payments-processors/lnurl-payments-processor.ts
im-adithya: As commented above, we only save the description/comment in our server but don't return it in the response, so it made more sense to have the pubkey only instead of the string with relay name and npub. Will check again!

src/repositories/invoice-repository.ts
im-adithya: Should I change the above ones as well along with this?

Cameri: oh yeah, deja vu! because `isNil` checks if the value is undefined, so the check is not needed

src/services/payments-service.ts
im-adithya: Does it? I had issues with it iirc

Idk if it's because of not including verify_url here? https://github.com/Cameri/nostream/pull/202/files/6bd859fd43f44f4db0cd3336ca5f7771cd6e5d29#diff-a3487de3d1ff9ff4f5025e4073c92c4f2d51c7b664a818b0fcd38fd5469f5674R127

im-adithya: (Testing again)

Cameri: It both updates and inserts. If it's failing with verify_url then it needs to be accounted for.
But I prefer explicitly mapping stuff as opposed to spreading an object because it's unclear what is being mapped.

im-adithya: Oops sorry, it's actually because we only have the information to update i.e. status (and id) coming from the getInvoice in the lnurl processor as we don't have access to the whole invoice by id from the server side (since we use a randomUUID)

im-adithya: https://github.com/Cameri/nostream/pull/202/files#diff-25cc4285e45b3c547df086050d23de6cca9038dde8a749713e83a28f78f229c7R24-R27

Cameri: @im-adithya should we just have an `updateInvoiceStatus` function?

im-adithya: Yes, that would be awesome!

Cameri: Yes, please! We might need a corresponding repository function too?

im-adithya: Done! I'm using the upsert by fetching the fullInvoice using Id as before

im-adithya: @Cameri just a gentle reminder on this one!

Cameri: You leaked your private keys. Please don't type private keys in the code again and always use environment variables to inject them.

im-adithya: Yeah had to remove them :') luckily didn't use them much
> always use environment variables to inject them.

Sure!

Cameri (approved):

Issue #206, opened 2023-02-15
[BUG] Expiration tag related issue


antonleviathan:
**Describe the bug**

**To Reproduce**
1. Run command '...'
2. See error

**Expected behavior**

**Screenshots**

**System (please complete the following information):**
 - OS: [e.g. Ubuntu]
 - Platform: [e.g. docker, standalone]
 - Version: [e.g. 1.9.11]

**Logs**
If applicable, attach log messages showing the bug and context when the bug happened.

**Additional context**
```
nostream          | {
nostream          |   kind: 4,
nostream          |   pubkey: '42b3db1ca9f73ea861cca1f5a9f74dadf97b6ff539cdf722ccae16119907dfe6',
nostream          |   tags: [
nostream          |     [
nostream          |       'p',
nostream          |       'cb46e92f5753ca1b330c59e98c5d88e1bbb1f93d4c0af918e6581f3091499175'
nostream          |     ],
nostream          |     [ 'expiration', '1676471110' ]
nostream          |   ],
nostream          |   content: '8/yASQYBb4mgIy+qlFGOml/VQkSTiO/bKPy1/Zpat8D11Ca7tQKFVgDyBIIuuWLw?iv=dhm9ggFCaE/VhyGXCg8QNw==',
nostream          |   created_at: 1676470810,
nostream          |   id: '7d5789510b0f2f0bbc12861ee33233f213488e69e9b702e72cfebd8ec6e60ff2',
nostream          |   sig: '1b17a53df1c91edd448c596a528b00d5edef9ce3de2d82a2485e6d7affe18fb5d8387a8e2ca001bc59632654249a069a909c52db984c972377034b8cc67e85e8',
nostream          |   [Symbol(Context)]: {
nostream          |     remoteAddress: SocketAddress { address: '138.197.76.21', port: 0, family: 'ipv4', flowlabel: 0 }
nostream          |   },
nostream          |   [Symbol(Expiration)]: 1676471110
nostream          | }


nostream          | ] TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received type number (1676471110)
nostream          |     at new NodeError (node:internal/errors:399:5)
nostream          |     at Function.from (node:buffer:334:9)
nostream          |     at toBuffer (/app/src/utils/transform.js:8:36)
nostream          |     at /app/node_modules/ramda/src/internal/_pipe.js:3:14
[...]

PR #207, opened 2023-02-15
fix: don't send expiration data to buff

antonleviathan:
## Related Issue
https://github.com/Cameri/nostream/issues/206


Cameri (approved):
LG

PR #271, opened 2023-03-28
fix: issues with invoices

Cameri:
## Types of changes

## Checklist:

### DIFF
--- a/resources/invoices.html
+++ b/resources/invoices.html
@@ -166,10 +166,7 @@
 
           switch (message[0]) {
             case 'EVENT': {
-              // TODO: validate event
               const event = message[2]
-              console.log('event', event)
-              // TODO: validate signature
               if (
                 event.pubkey === relayPubkey
                 && event.kind === 402
--- a/resources/invoices.html
+++ b/resources/invoices.html
@@ -107,8 +107,11 @@
       var timeout
       var paid = false
       var fallbackTimeout
+      var now = Math.floor(Date.now()/1000)
 
       console.log('invoice id', reference)
+      console.log('pubkey', pubkey)
+      console.log('bolt11', invoice)
 
       function getBackoffTime() {
         return 5000 + Math.floor(Math.random() * 5000)
@@ -149,7 +152,8 @@
         var socket = new WebSocket(relayUrl)
         socket.onopen = () => {
           console.log('connected')
-          socket.send(JSON.stringify(['REQ', 'payment', { kinds: [4], authors: [relayPubkey], '#c': [reference], limit: 1 }]))
+          var subscription = ['REQ', 'payment', { kinds: [402], '#p': [pubkey], since: now - 60 }]
+          socket.send(JSON.stringify(subscription))
         }
 
         socket.onmessage = (raw) => {
@@ -164,14 +168,23 @@
             case 'EVENT': {
               // TODO: validate event
               const event = message[2]
+              console.log('event', event)
               // TODO: validate signature
-              if (event.pubkey === relayPubkey) {
-                paid = true
+              if (
+                event.pubkey === relayPubkey
+                && event.kind === 402
+              ) {
+                const pubkeyTag = event.tags.find((t) => t[0] === 'p' && t[1] === pubkey)
+                const invoiceTag = event.tags.find((t) => t[0] === 'bolt11' && t[1] === invoice)
 
-                if (expiresAt) clearTimeout(timeout)
+                if (pubkeyTag && invoiceTag) {
+                  paid = true
 
-                hide('pending')
-                show('paid')
+                  if (expiresAt) clearTimeout(timeout)
+
+                  hide('pending')
+                  show('paid')
+                }
               }
             }
             break;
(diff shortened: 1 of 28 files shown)


resources/invoices.html
```diff
@@ -164,14 +168,23 @@ <h2 class="text-danger">Invoice expired!</h2>
             case 'EVENT': {
               // TODO: validate event
               const event = message[2]
+              console.log('event', event)
```
antonleviathan, line 171: Do we need this log?

resources/invoices.html
```diff
@@ -164,14 +168,23 @@ <h2 class="text-danger">Invoice expired!</h2>
             case 'EVENT': {
               // TODO: validate event
```
antonleviathan, line 169: Is this an old comment?

src/app/maintenance-worker.ts
```diff
@@ -47,21 +47,26 @@ export class MaintenanceWorker implements IRunnable {
...
           invoice.status !== updatedInvoice.status
           && updatedInvoice.status == InvoiceStatus.COMPLETED
-          && invoice.confirmedAt
+          && updatedInvoice.confirmedAt
         ) {
           debug('confirming invoice %s & notifying %s', invoice.id, invoice.pubkey)
+          const update = {
+              ...invoice,
+              ...{ amountPaid: invoice.amountRequested },
+              ...updatedInvoice,
```
antonleviathan, line 65: Should we use a [`deepMerge`](https://ramdajs.com/docs/#mergeDeepLeft) here?

Cameri: done

resources/invoices.html
Cameri: indeed

src/handlers/event-message-handler.ts
```diff
@@ -249,6 +261,10 @@ export class EventMessageHandler implements IMessageHandler {
       return
     }
 
+    if (this.getRelayPublicKey() === event.pubkey) {
+      return undefined
```
antonleviathan, line 265: Inconsistent returns, is this intentional?

src/services/payments-service.ts
```diff
@@ -309,31 +245,33 @@ ${invoice.bolt11}`,
...
     }
 
-    const persistEvent = this.eventRepository.create.bind(this.eventRepository)
+    const persistEvent = async (event: Event) => {
+      await this.eventRepository.create(event)
+
+      return event
+    }
 
     await pipe(
```
antonleviathan, line 270: Should we be catching here for errors?

src/handlers/event-message-handler.ts
Cameri: this is fine, we return `undefined` if user is admitted/whitelisted or a string with the reason if they are not admitted

this new check fixes the issue of the relay's public key not being allowed on the relay when payment is enabled which is a big oversight

src/utils/event.ts
```diff
@@ -181,15 +181,33 @@ export const identifyEvent = async (event: UnidentifiedEvent): Promise<UnsignedE
   return { ...event, id }
 }
 
-export function getRelayPrivateKey(relayUrl: string): string {
+let privateKeyCache: string | undefined
+export function getRelayPrivateKey(secret: string): string {
```
antonleviathan, line 185: Is `secret` an optional parameter? If so we could make the type indicate this

src/utils/event.ts
```diff
@@ -181,15 +181,33 @@ export const identifyEvent = async (event: UnidentifiedEvent): Promise<UnsignedE
...
 }
 
-export const getPublicKey = (privkey: string | Buffer) => Buffer.from(secp256k1.getPublicKey(privkey, true)).subarray(1).toString('hex')
+const publicKeyCache: Record<string, string> = {}
+export const getPublicKey = (privkey: string) => {
+  if (privkey in publicKeyCache) {
+    return publicKeyCache[privkey]
+  }
+
+  publicKeyCache[privkey] = Buffer.from(secp256k1.getPublicKey(privkey, true)).subarray(1).toString('hex')
```
antonleviathan, line 207: You can do something like this instead:
`secp256k1.utils.bytesToHex(secp256k1.getPublicKey(privkey, true)).subarray(1))`

src/utils/event.ts
Cameri: it should always be the relay's URL, but it is ignored when `RELAY_PRIVATE_KEY` env var is set

src/handlers/event-message-handler.ts
Cameri: in the future we could change this pattern to using constants or symbols

src/utils/event.ts
antonleviathan: Ok, why is the param called `secret`?

src/services/payments-service.ts
Cameri: sure! will add a catch

src/handlers/event-message-handler.ts
Cameri: Ah, I'll remove the extra undefined

Issue #381, opened 2025-02-12
[BUG] Nip-11 seems to be served on root domain, not on relay_url


MegalithicBTC:
Feel free to close this if I'm confused.

The NIP-11 spec (https://github.com/nostr-protocol/nips/blob/master/11.md#relay-information-document ) 
... says that...

```
Relays may provide server metadata... This is made available as a JSON document over HTTP, on the same URI as the relay's websocket. )

```

Operative text here is "on the same URI as the relay's websocket"

So I THINK that means if my `relay_url` is set to...

```
wss://mydomain.com/v1
```
Then the Nip-11 data should be accessible like ....

```
curl -i   -H "Accept: application/nostr+json"   https://some-domain.com/v1
```
But so far I am finding this not to be the case with Nostream. The Nip-11 data is instead accessible on the ROOT of the domain

```
curl -i   -H "Accept: application/nostr+json"   https://some-domain.com
```

theAnuragMishra:
yeah should be available on `/v1`.

cameri:
Thank you @sagar-h007 !

PR #399, opened 2026-04-08
fix(router): NIP-11 being served only on root instead of relay path

sagar-h007:
## Description
This PR fixes an issue where the NIP-11 relay information document was only being served at the root (`/`) of the domain, even when the relay itself was configured on a different path (for example `/v1`). 

Specifically, this PR introduces the following changes:
* Adds a top-level request interceptor that checks for `Accept: application/nostr+json`.
* Routes matching requests to the NIP-11 handler regardless of the path.
* Updates the NIP-11 handler to use flexible header matching (`includes` instead of strict equality).
* Normalizes header checks to avoid issues with compound or case variations.

## Related Issue
#381 

## Motivation and Context
According to the NIP-11 spec, the relay metadata should be available on the same URI that handles the WebSocket connection. In practice, this means a relay at `wss://domain.com/v1` should expose its NIP-11 document at `https://domain.com/v1`, not just at `https://domain.com/`.

Previously, WebSocket upgrades in Nostream were effectively path-agnostic, while Express routes were explicitly bound to `/`. As a result, requests like `GET /v1` with `Accept: application/nostr+json` returned a `404` instead of the expected NIP-11 response. This change keeps the fix minimal while aligning behavior more closely with the NIP-11 spec and how relays are typically exposed behind proxies or path prefixes.

## How Has This Been Tested?
Tested locally by verifying that NIP-11 is now correctly served on the relay path (e.g., `/v1`). Verified that the existing behavior on the root (`/`) remains unchanged and that the updated handler successfully processes requests from real-world clients sending compound Accept headers.

**Test run:**
```bash
curl -H "Accept: application/nostr+json" http://localhost:8008/v1
```

## Types of changes

## Checklist:

### DIFF
--- a/src/handlers/request-handlers/root-request-handler.ts
+++ b/src/handlers/request-handlers/root-request-handler.ts
@@ -1,4 +1,5 @@
 import { NextFunction, Request, Response } from 'express'
+import accepts from 'accepts'
 import { path } from 'ramda'
 
 import { createSettings } from '../../factories/settings-factory'
@@ -9,7 +10,7 @@ import packageJson from '../../../package.json'
 export const rootRequestHandler = (request: Request, response: Response, next: NextFunction) => {
   const settings = createSettings()
 
-  if (request.header('accept')?.includes('application/nostr+json')) {
+  if (accepts(request).type(['application/nostr+json'])) {
     const {
       info: { name, description, pubkey: rawPubkey, contact, relay_url },
     } = settings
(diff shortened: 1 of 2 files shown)


cameri:
Do you mind rebasing your PR on top of the main branch?

cameri (approved):

cameri:
@sagar-h007 Please fix the linting issues.

Issue #359, opened 2024-01-30
[REQUEST] Purging old events


AbelLykens:
To keep the node managable (in terms of resources, lookups, indices etc) it would be nice to be able to delete events after _x_ days.

I have considered doing manual queries, but that did not feel like a good idea.

Justxd22:
i will add a config for it and default = 30 days

cameri:
thank you @Justxd22 !

PR #412, opened 2026-04-09
feat: add opt-in event retention purge (#359)

Justxd22:
- Added `limits.event.retentionDays` to `settings.yaml`.
- purge expired, soft-deleted, and older-than-retention events from maintenance worker loop
- Added efficient batch-delete query in EventRepository that handles retention, NIP-40 (expiration), and NIP-09 (soft-deletes) in a single operation.
- keep behavior opt-in: no purge when `retentionDays` is unset/non-positive

## Description
The logic is hooked into the existing heartbeat of the MaintenanceWorker:
   1. Every 60 seconds, the worker wakes up to check for pending invoices.
   2. It now also checks the retentionDays setting.
   3. If enabled, it executes a single SQL DELETE query targeting:
       - Events where created_at is older than the retention threshold.
       - Events with an expired expires_at timestamp (NIP-40).
       - Events marked as deleted via deleted_at (NIP-09).
   4. Results are logged to the console using [Maintenance] tags for easy monitoring.

Add the following to your `.nostr/settings.yaml`:
```yml
limits:
    event:
        retentionDays: 30 # Set to 0 or comment out to disable.
```

## Related Issue
#359 
> @Justxd22 Would be good to have a setting so this is off if a date is not set. Making it default to 30 day would be a breaking change.

## How Has This Been Tested?
- i made test script that add events using wscat with timestamp older than 30 days => wait for next 60 sec and check logs
- i disabled the 30 days config and used same script => no delete
- re-added 30 days config and restarted => on startup it checked older events and auto deleted them

image

## Types of changes

## Checklist:

### DIFF
--- a/resources/default-settings.yaml
+++ b/resources/default-settings.yaml
@@ -76,6 +76,13 @@ limits:
       - "10.10.10.1"
       - "::ffff:10.10.10.1"
   event:
+    retention:
+      maxDays: -1
+      kind:
+        whitelist:
+          - 62
+      pubkey:
+        whitelist: []
     eventId:
       minLeadingZeroBits: 0
     kind:
--- a/src/repositories/event-repository.ts
+++ b/src/repositories/event-repository.ts
@@ -31,7 +31,7 @@ import {
 import { ContextMetadataKey, EventDeduplicationMetadataKey, EventExpirationTimeMetadataKey, EventKinds } from '../constants/base'
 import { DatabaseClient, EventId } from '../@types/base'
 import { DBEvent, Event } from '../@types/event'
-import { IEventRepository, IQueryResult } from '../@types/repositories'
+import { EventPurgeCounts, EventRetentionOptions, IEventRepository, IQueryResult } from '../@types/repositories'
 import { toBuffer, toJSON } from '../utils/transform'
 import { createLogger } from '../factories/logger-factory'
 import { isGenericTagQuery } from '../utils/filter'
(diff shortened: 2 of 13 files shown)


cameri (changes requested):
I think some changes are required to move forward, but mostly looking pretty good!

src/repositories/event-repository.ts
```diff
@@ -251,4 +251,27 @@ export class EventRepository implements IEventRepository {
...
+
+    debug('deleting expired and retained events (retentionLimit: %d, now: %d)', retentionLimit, now)
+
+    return this.masterDbClient('events')
+      .where(function () {
+        this.where(function () {
+          this.whereNotNull('expires_at').andWhere('expires_at', '<', now)
+        })
+        .orWhereNotNull('deleted_at')
+        .orWhere('event_created_at', '<', retentionLimit)
```
cameri: We are matching events based on the event's created_at, expires_at (if set), and deleted_at (if set), but the documentation in CONFIGURATION.md only mentions created_at.

resources/default-settings.yaml
```diff
@@ -76,6 +76,7 @@ limits:
       - "10.10.10.1"
       - "::ffff:10.10.10.1"
   event:
+    retentionDays: -1
```
cameri: After thinking about this a little more, we will definitely need `kind` and `pubkey` whitelists.

Example:
```
event:
  retention:
    maxDays: -1
    kind:
      whitelist: [62]
    pubkey:
      whitelist: []
```

Why: We recently merged https://github.com/cameri/nostream/pull/418 which creates tombstones for users requesting to vanish. If we delete these tombstones NIP-62 support will be ineffective if retention days is enabled.

Why: The owner of a relay may want to only retain events for their own pubkey and their friends, and delete everyone else's. This one can be a separate ticket.

src/repositories/event-repository.ts
```diff
@@ -251,4 +251,27 @@ export class EventRepository implements IEventRepository {
...
+    }
+
+    const retentionLimit = now - (retentionDays * 86400)
+
+    debug('deleting expired and retained events (retentionLimit: %d, now: %d)', retentionLimit, now)
+
+    return this.masterDbClient('events')
+      .where(function () {
+        this.where(function () {
+          this.whereNotNull('expires_at').andWhere('expires_at', '<', now)
```
cameri: I think that `whereNotNull('expires_at')` is redundant because andWhere('expires_at', '<', now) would not match events where expires_at is NULL.

cameri:
Looking good! Left just a few comments.

src/repositories/event-repository.ts
```diff
@@ -251,4 +251,27 @@ export class EventRepository implements IEventRepository {
...
+        })
+        .orWhereNotNull('deleted_at')
+        .orWhere('event_created_at', '<', retentionLimit)
+      })
+      .del()
```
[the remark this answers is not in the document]
cameri: Good point. Maybe we can just limit deletion to 1000 events at a time? Depending on how often the maintenance worker runs the scheduled job it will eventually purge all

cameri:
Please address Copilot comments

cameri (approved):

PR #423, opened 2026-04-09
feat: add nginx reverse proxy to docker-compose

### LINKED ISSUE #36: Add nginx to docker-compose template
In order to support NIP-05, certificates and possibly other features separate from the relay (Admin page), need to add an HTTP proxy that will sit in between those services and the clients

https://hub.docker.com/_/nginx/

kanishka0411:
## Description
Added an nginx reverse proxy with Let's Encrypt SSL as an optional docker-compose overlay. It follows the same pattern as the tor setup-separate compose file, separate start script, doesn't touch the main start flow.

## Related Issue
Fixes #36 

## Motivation and Context
Operators need a reverse proxy for TLS, NIP-05, etc. This gives a ready-to-go nginx + certbot setup without getting in the way of people who already run their own proxy.

## How Has This Been Tested?
* Script syntax validated with bash -n
* Compose config merge validated with docker compose config
* Full ACME flow needs a public domain so couldn't test that locally, would appreciate help testing if possible

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,10 @@ Start:
   ```
   ./scripts/start_with_tor
   ```
+  or, with Nginx reverse proxy and Let's Encrypt SSL:
+  ```
+  RELAY_DOMAIN=relay.example.com CERTBOT_EMAIL=you@example.com ./scripts/start_with_nginx
+  ```
 
 Stop the server with:
   ```
--- a/docker-compose.nginx.yml
+++ b/docker-compose.nginx.yml
@@ -1,0 +1,63 @@
+services:
+  nginx:
+    image: nginx:1.25-alpine
+    container_name: nostream-nginx
+    ports:
+      - 80:80
+      - 443:443
+    volumes:
+      - ${PWD}/nginx/conf.d:/etc/nginx/conf.d
+      - ${PWD}/nginx/ssl:/etc/nginx/ssl
+      - certbot-webroot:/var/www/certbot
+    depends_on:
+      - nostream
+    restart: on-failure
+    # Run nginx in foreground (so container exits if nginx dies).
+    # A background loop watches for a signal file created by certbot
+    # after cert issuance/renewal, and reloads nginx within seconds.
+    command: >
+      /bin/sh -c "while :; do
+        if [ -f /etc/nginx/ssl/reload-nginx ]; then
+          nginx -s reload;
+          rm -f /etc/nginx/ssl/reload-nginx;
+        fi;
+        sleep 5;
+      done & nginx -g 'daemon off;'"
+    networks:
+      default:
+
+  certbot:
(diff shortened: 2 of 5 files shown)


docker-compose.nginx.yml
```diff
@@ -0,0 +1,63 @@
...
+    networks:
+      default:
+
+  certbot:
+    image: certbot/certbot:latest
+    container_name: nostream-certbot
+    volumes:
+      - ${PWD}/nginx/ssl:/etc/letsencrypt
+      - certbot-webroot:/var/www/certbot
+    entrypoint: >
```
cameri: should this be an entrypoint script instead so that we can keep the compose file neat and tidy?

README.md
```diff
@@ -209,16 +209,10 @@ Start:
...
-**Windows / WSL2 users:** Docker bind-mounts can cause PostgreSQL permission errors on Windows. Use the dedicated override file instead:
-  ```
-  docker compose -f docker-compose.yml -f docker-compose.windows.yml up --build
-  ```
-  Or add this to your `.env` file so you don't have to type it every time:
+  or, with Nginx reverse proxy and Let's Encrypt SSL:
   ```
-  COMPOSE_FILE=docker-compose.yml:docker-compose.windows.yml
+  RELAY_DOMAIN=relay.example.com CERTBOT_EMAIL=you@example.com ./scripts/start_with_nginx
   ```
```
cameri, line 215: 👏

docker-compose.nginx.yml
kanishka0411: Done! Moved it to ```scripts/certbot_entrypoint.sh``` :-)

docker-compose.nginx.yml
```diff
@@ -0,0 +1,48 @@
...
+    image: certbot/certbot:v2.11.0
+    container_name: nostream-certbot
+    environment:
+      RELAY_DOMAIN: ${RELAY_DOMAIN:?RELAY_DOMAIN required}
+      CERTBOT_EMAIL: ${CERTBOT_EMAIL:?CERTBOT_EMAIL required}
+    volumes:
+      - ${PWD}/nginx/ssl:/etc/letsencrypt
+      - ${PWD}/scripts/certbot_entrypoint.sh:/entrypoint.sh:ro
+      - certbot-webroot:/var/www/certbot
+    entrypoint: /entrypoint.sh
```
cameri, line 40: The path seems incorrect.

kanishka0411: Ah, which path are you referring to,the host side (scripts/certbot_entrypoint.sh) or the mount inside the container (/entrypoint.sh)? Want me to move it somewhere else?

cameri: Never mind, I misunderstood the code.

Issue #441, opened 2026-04-10
upgrade express to 4.22.1 to resolve high-severity vulnerabilities


Mahmoud-s-Khedr:
## Summary

Several high-severity vulnerabilities exist in the current `express@4.18.2` installation through its transitive dependencies. Upgrading to `express@4.22.1` resolves all of them in a single patch.

## Affected packages

| Package | Vulnerable range | Severity | Advisory |
|---|---|---|---|
| `express` | <=4.21.2 | High | Depends on all packages below |
| `body-parser` | <=1.20.3 | High | DoS via URL encoding ([GHSA-qwcr-r2fm-qrc7](https://github.com/advisories/GHSA-qwcr-r2fm-qrc7)) |
| `path-to-regexp` | <=0.1.12 | High | ReDoS backtracking ([GHSA-9wv6-86v2-598j](https://github.com/advisories/GHSA-9wv6-86v2-598j), [GHSA-rhx6-c78j-4q9w](https://github.com/advisories/GHSA-rhx6-c78j-4q9w), [GHSA-37ch-88jc-xwx2](https://github.com/advisories/GHSA-37ch-88jc-xwx2)) |
| `send` | <0.19.0 | — | Template injection → XSS ([GHSA-m6fv-jmcg-4jfg](https://github.com/advisories/GHSA-m6fv-jmcg-4jfg)) |
| `serve-static` | <=1.16.0 | — | Depends on vulnerable `send` |
| `cookie` | <0.7.0 | — | Out-of-bounds chars in cookie name/path/domain ([GHSA-pxg6-pf52-xh8x](https://github.com/advisories/GHSA-pxg6-pf52-xh8x)) |
| `qs` | <=6.14.1 | Moderate | DoS via arrayLimit bypass ([GHSA-w7fw-mjwx-w883](https://github.com/advisories/GHSA-w7fw-mjwx-w883), [GHSA-6rw7-vpxm-498p](https://github.com/advisories/GHSA-6rw7-vpxm-498p)) |

PR #442, opened 2026-04-10
chore: update dependencies for express, body-parser, and js-yaml

Mahmoud-s-Khedr:
## Description
Updated the following dependencies to their latest patch/minor versions to address security vulnerabilities:
- `express`: `4.18.2` → `^4.22.1`
- `body-parser`: `1.20.1` → `^1.20.4`
- `js-yaml`: `4.1.0` → `^4.1.1`
- `@types/express`: `4.17.15` → `^4.17.21`

## Related Issue
#441 

## Motivation and Context
These dependency updates address known security vulnerabilities in older versions of `express` and `body-parser`. Keeping dependencies up to date reduces the attack surface and ensures the project benefits from upstream bug fixes and security patches. The updates are backwards-compatible (patch/minor semver bumps) and should not introduce breaking changes.

## How Has This Been Tested?
- Ran `npm install` to regenerate `package-lock.json` with the updated versions.
- Verified the application starts correctly with the updated dependencies.
- Existing test suite was run to confirm no regressions.

## Types of changes

## Checklist:


package.json
```diff
@@ -83,7 +83,7 @@
     "@types/chai": "^4.3.1",
     "@types/chai-as-promised": "^7.1.5",
     "@types/debug": "4.1.7",
-    "@types/express": "4.17.15",
+    "@types/express": "^4.17.21",
```
cameri: Can we stick with pinned versions instead here and in other places? I prefer to avoid the ^ so we don't end up installing a malicious package.

cameri (changes requested):
Let's make sure we pin versions.

package.json
Mahmoud-s-Khedr: the packages updated in this PR are now pinned

cameri:
There's merge conflicts now. Do you mind fixing those?

PR #454, opened 2026-04-11
fix: OpenNode callback accepts unauthenticated requests

### LINKED ISSUE #453: [BUG] OpenNode callback accepts unauthenticated requests and never parses the webhook body
**Describe the bug**

The `/callbacks/opennode` accepts any incoming request without verifying its authenticity. Every other callback handler (Zebedee, LNbits, Nodeless) validates the source , OpenNode is the only one that doesn't.

This has two problems:

1. **No HMAC check.** OpenNode signs every webhook with a `hashed_order` field (`HMAC-SHA256(api_key, charge_id)`). The endpoint ignores it entirely.

2. **Wrong body parser.** OpenNode sends webhooks as `application/x-www-form-urlencoded` (per their docs), but the route registers `json()` middleware only. The body is silently unparsed for real OpenNode requests, so even if validation were added, it would always operate on an empty object.

An attacker with a valid invoice ID can POST a fake `status: paid` payload and permanently corrupt that invoice's state in the database, making it uncompletable by the real payer.

**To Reproduce**

1. Stand up the relay with OpenNode as the payment processor.
2. Grab a valid invoice ID from the connection/invoice flow.
3. Run:
```bash
curl -X POST https://<your-relay>/callbacks/opennode \
  --data "id=<invoice_id>&status=paid&order_id=<any_pubkey>&hashed_order=fakehash"
```
4. Check the `invoices` table , `status` is now `COMPLETED` with no payment made.

**Expected behavior**

The endpoint should parse `application/x-www-form-urlencoded` bodies (as OpenNode sends them) and reject any request where `hashed_order` does not match `HMAC-SHA256(OPENNODE_API_KEY, charge_id)`, which is what OpenNode's own documentation specifies and what the other handlers already do for their respective authentication schemes.

**Screenshots**

N/A

Anshumancanrock:
## Description

Two bugs in the OpenNode callback handler

1. **No signature verification.** The controller accepted any request unconditionally. An attacker could POST to `/callbacks/opennode` and flip an invoice's status to `COMPLETED` in the database.

2. **Body was never parsed.** OpenNode sends `application/x-www-form-urlencoded` payloads, but the route only registered `json()` middleware. So `request.body` was always `{}` , the handler was running on empty input every time.

This PR fixes both.

Closes #453 

## Testing

Added unit tests covering:
-  wrong processor will throw 403
-  missing/malformed body fields will thow 400
-  invalid signature will thow 403
- valid signed request = invoice processed

And a route-level test confirming the form-encoded body is actually parsed before hitting the controller.

## Types of changes

## Checklist


Anshumancanrock:
hi @cameri , could you please review this PR? Thank you !

cameri:
@copilot resolve the merge conflicts in this pull request

Anshumancanrock:
hii @cameri , I’ve resolved the merge conflicts. Please take another look when you get a chance.

src/controllers/callbacks/opennode-callback-controller.ts
```diff
@@ -20,19 +22,92 @@ export class OpenNodeCallbackController implements IController {
...
+      response
+        .status(403)
+        .send('Forbidden')
+      return
+    }
+
+    const validStatuses = ['expired', 'refunded', 'unpaid', 'processing', 'underpaid', 'paid']
+
+    if (
+      !request.body
```
cameri: Can we validate the request using zod?

src/controllers/callbacks/opennode-callback-controller.ts
```diff
@@ -20,19 +22,92 @@ export class OpenNodeCallbackController implements IController {
     response: Response,
   ) {
     debug('request headers: %o', request.headers)
-    debug('request body: %O', request.body)
+    debug(
+      'request body metadata: hasId=%s hasHashedOrder=%s status=%s',
+      typeof request.body?.id === 'string',
+      typeof request.body?.hashed_order === 'string',
+      typeof request.body?.status === 'string' ? request.body.status : 'missing',
```
cameri: we should log this after we've validated the request

cameri:
@Anshumancanrock some conflicts maybe with the other PR

PR #470, opened 2026-04-14
test: add missing unit tests for message factory functions

d1vyanshu-kumar:
### Summary

This PR adds missing unit coverage for message factory functions in `src/utils/messages.ts`.

### What changed

Added tests in `test/unit/utils/messages.spec.ts` for:

* `createCommandResult`
  Covers both success and rejection cases.

* `createSubscriptionMessage`
  Covers single-filter and multi-filter cases.

* `createRelayedEventMessage`
  Covers both paths: with and without a secret.

### Why

These functions are exported and used by the relay, but they previously had missing test coverage. Adding these tests helps catch regressions in message creation logic before they reach production.

### Verification

* `npm run test:unit`
* All unit tests pass


d1vyanshu-kumar:
@cameri Could you please review this PR and let me know if it looks good?

cameri (approved):

PR #493, opened 2026-04-18
fix: only trust forwarded IP header from configured trusted proxies

### LINKED ISSUE #492: IP spoofing via untrusted x-forwarded-for header
**Describe the bug**
Nostream trusts the x-forwarded-for header from any client without checking if the request actually came from a legitimate proxy. This allows an attacker to fake their IP address and bypass rate limits, IP blacklists, and payment throttling.

**To Reproduce**
1. Configure network.remoteIpHeader: x-forwarded-for (this is the default)
2. Connect directly to Nostream's port (bypassing your proxy)
3. Set header x-forwarded-for: <any IP> in your request
4. Nostream treats you as that IP -rate limits and IP controls are bypassed

**Expected behavior**
The forwarding header should only be trusted when the request comes from a known proxy. Direct clients should always use the real socket address.

**Screenshots**

OS: Any
Platform: Docker / Standalone
Version: All current versions 

**Logs**
If applicable, attach log messages showing the bug and context when the bug happened.

**Additional context**

kanishka0411:
## Description
Added network.trustedProxies config option. The x-forwarded-for header (or any configured remoteIpHeader) is now only trusted when the request's socket IP is in the trustedProxies list. Otherwise Nostream falls back to the real socket address. Also added a runtime warning when remoteIpHeader is set but trustedProxies is empty.

## Related Issue
Fixes #492 

## Motivation and Context
Previously any client could set x-forwarded-for to any IP and Nostream would blindly trust it. This allowed attackers to spoof their IP and bypass rate limits, IP blacklists, and payment throttling.

## How Has This Been Tested?
Added unit tests for:
- trusted proxy allows forwarded header
- untrusted proxy falls back to socket address
- IPv4-mapped proxy address normalization
  
## Types of changes

## Checklist:

### DIFF
--- a/resources/default-settings.yaml
+++ b/resources/default-settings.yaml
@@ -57,6 +57,8 @@ network:
   maxPayloadSize: 524288
   # Comment the next line if using CloudFlare proxy
   remoteIpHeader: x-forwarded-for
+  # Optional: only trust forwarding headers from these proxy IPs
+  trustedProxies: []
   # Uncomment the next line if using CloudFlare proxy
   # remoteIpHeader: cf-connecting-ip
 workers:
--- a/resources/default-settings.yaml
+++ b/resources/default-settings.yaml
@@ -55,12 +55,16 @@ nip05:
   domainBlacklist: []
 network:
   maxPayloadSize: 524288
-  # Comment the next line if using CloudFlare proxy
-  remoteIpHeader: x-forwarded-for
-  # Optional: only trust forwarding headers from these proxy IPs
-  trustedProxies: []
-  # Uncomment the next line if using CloudFlare proxy
+  # Uncomment only when using a trusted reverse proxy and configuring trustedProxies.
+  # remoteIpHeader: x-forwarded-for
   # remoteIpHeader: cf-connecting-ip
+  # Proxy IPs allowed to set remoteIpHeader (loopback and common docker internal)
+  trustedProxies:
+    - "127.0.0.1"
+    - "::ffff:127.0.0.1"
+    - "::1"
+    - "10.10.10.1"
+    - "::ffff:10.10.10.1"
 workers:
   count: 0
 mirroring:
--- a/src/utils/http.ts
+++ b/src/utils/http.ts
@@ -5,6 +5,28 @@ import { Settings } from '../@types/settings'
 
 const logger = createLogger('http-utils')
 
+const normalizeIpAddress = (input: string): string => {
+  if (input.startsWith('::ffff:')) {
+    return input.slice(7)
+  }
+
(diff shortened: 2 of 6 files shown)


src/utils/http.ts
```diff
@@ -2,6 +2,28 @@ import { IncomingMessage } from 'http'
 
 import { Settings } from '../@types/settings'
 
+const normalizeIpAddress = (input: string): string => {
+  if (input.startsWith('::ffff:')) {
```
cameri, line 9: Should we use Node built-in Net module instead?

https://nodejs.org/api/net.html#socketaddressparseinput

resources/default-settings.yaml
```diff
@@ -39,6 +39,8 @@ network:
   maxPayloadSize: 524288
   # Comment the next line if using CloudFlare proxy
   remoteIpHeader: x-forwarded-for
```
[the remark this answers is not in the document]
cameri: Can we add common loop back / docker internal IPs here? There may be examples in the settings file

src/utils/http.ts
kanishka0411: this branch is already on Node 24 (.nvmrc v24.14.1, @types/node ^24.0.0), so SocketAddress.parse() should be available without a cast. I can switch to that if you prefer, or keep the current explicit normalization.

cameri: I'd prefer if we switch to the Node built in implementation if it works for our purpose. Less code is better code.

kanishka0411: I tried switching to `SocketAddress.parse()`, but it returns `undefined` for raw forms we see in `remoteAddress` like `::ffff:127.0.0.1` and `::1`, so it doesn’t fully normalize this path by itself.I can update it to use `SocketAddress.parse()` first and fall back to the current `::ffff:` normalization when parsing is unsupported. Does that work for you?

resources/default-settings.yaml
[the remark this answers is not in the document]
kanishka0411: Sure, will add those :-)

cameri:
@kanishka0411 looks like we have some conflicts now. I'm wondering if we should re-order CONFIGURATION.md list of settings A-Z so there's less conflicts.

src/utils/http.ts
cameri: no, let's keep it as-is for now. thanks for looking into it

cameri:
@copilot resolve the merge conflicts in this pull request

cameri:
@copilot resolve the merge conflicts in this pull request

cameri (approved):

Issue #541, opened 2026-04-19
[BUG] Redis auth attempted with unset credentials when REDIS_URI is absent


Justxd22:
**Describe the bug**
Nostream attempts Redis authentication even when no Redis credentials are configured, causing WRONGPASS on local Redis instances that allow unauthenticated access. 

`src/cache/client.ts`:
```ts
export const getCacheConfig = (): RedisClientOptions => ({
  url: process.env.REDIS_URI
    ? process.env.REDIS_URI
    : `redis://${process.env.REDIS_USER}:${process.env.REDIS_PASSWORD}@${process.env.REDIS_HOST}:${process.env.REDIS_PORT}`,
  password: process.env.REDIS_PASSWORD,
})
```
notice how it uses `process.env.REDIS_PASSWORD` even when it's empty/unset or null, resulting in failed connection due to broken URI

**To Reproduce**
1. Configure .env with only:  
```ts
REDIS_HOST=localhost
REDIS_PORT=6379
```
2. Start nostream `pnpm run dev` and trigger a redis operation like visiting `127.0.0.1:8008/invoices`
3. Observe repeated Redis connection/auth errors.   

**Expected behavior**
If credentials `REDIS_PASSWORD` are unset, nostream should connect to Redis without AUTH.

**System (please complete the following information):**
 - OS: Arch linux
 - Platform: pnpm
 - Version: latest

**Logs**
```ts
uncaught error: [ErrorReply: WRONGPASS invalid username-password pair or user is disabled.]
uncaught error: [ErrorReply: WRONGPASS invalid username-password pair or user is disabled.]
uncaught error: [ErrorReply: WRONGPASS invalid username-password pair or user is disabled.]
uncaught error: [ErrorReply: WRONGPASS invalid username-password pair or user is disabled.]
uncaught error: [ErrorReply: WRONGPASS invalid username-password pair or user is disabled.]
uncaught error: [ErrorReply: WRONGPASS invalid username-password pair or user is disabled.]
uncaught error: [ErrorReply: WRONGPASS invalid username-password pair or user is disabled.]
uncaught error: [ErrorReply: WRONGPASS invalid username-password pair or user is disabled.]
uncaught error: [ErrorReply: WRONGPASS invalid username-password pair or user is disabled.]
```

cameri:
Thank you @Justxd22 !

PR #545, opened 2026-04-19
fix: skip Redis auth when credentials are unset with unit test (#541)

Justxd22:
## Description
  - fix Redis cache config to skip AUTH when `REDIS_PASSWORD` is unset
  - use `redis://host:port` when no credentials are provided          
  - keep `REDIS_URI` precedence when present                
  - add unit test coverage in `test/unit/cache/client.spec.ts` for:                                                                                                   
    - no `REDIS_URI` + no password                                                                                                                                    
    - password-based config                                                                                                                                           
    - `REDIS_URI` precedence 
  
## Related Issue
#541

## How Has This Been Tested?
`npm run test:unit`

## Types of changes

## Checklist:


src/cache/client.ts
```diff
@@ -4,12 +4,49 @@ import { createLogger } from '../factories/logger-factory'
...
+
+export const getCacheConfig = (): RedisClientOptions => {
+  if (process.env.REDIS_URI) {
+    return {
+      url: process.env.REDIS_URI,
+      ...(process.env.REDIS_PASSWORD ? { password: process.env.REDIS_PASSWORD } : {}),
+    }
+  }
+
+  const hasPassword = Boolean(process.env.REDIS_PASSWORD)
```
cameri: this condition is checked twice, here and in line 28

src/cache/client.ts
```diff
@@ -4,12 +4,49 @@ import { createLogger } from '../factories/logger-factory'
...
+  const host = process.env.REDIS_HOST
+  const port = process.env.REDIS_PORT
+
+  if (hasPassword) {
+    const username = process.env.REDIS_USER ?? 'default'
```
cameri, line 38: Is `default` in the Redis documentation? how does this default user come to be?

Justxd22: yes by default it's 'default' 

> https://redis.io/docs/latest/operate/rs/security/access-control/manage-users/default-user/
> https://redis.io/docs/latest/operate/oss_and_stack/management/security/acl

Issue #532, opened 2026-04-19
[BUG] `/` root in browser returns json


Justxd22:
**Describe the bug**
PR #399 introduce serving NIP-11 on root when headers contain `Accept: application/nostr+json` but the detection logic is failing for browsers and returning json

headers sent:
```bash
GET / HTTP/1.1
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8
Accept-Encoding: gzip, deflate, br, zstd
Accept-Language: en-US,en;q=0.8
Cache-Control: max-age=0
Connection: keep-alive
Host: 127.0.0.1:8008
If-None-Match: W/"2cf-yiYUGZLRa1VnCsMoAYdocVUDb08"
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: none
Sec-Fetch-User: ?1
Sec-GPC: 1
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36
sec-ch-ua: "Chromium";v="146", "Not-A.Brand";v="24", "Brave";v="146"
sec-ch-ua-mobile: ?0
sec-ch-ua-platform: "Linux"
```

**To Reproduce**
visit `127.0.0.1:8008` in browser

**Expected behavior**
returns html

**Screenshots**
<img width="589" height="339" alt="Image" src="https://github.com/user-attachments/assets/3515dfff-ffd9-4236-aa6d-d5f17615e31d" />

**System (please complete the following information):**
 - OS: arch linux
 - Platform: pnpm
 - Version: latest

Justxd22:
merged #546

PR #539, opened 2026-04-19
feat: add NWC payments processor (#323)

Justxd22:
Add new alby payments processor using NIP-47 NWC (make_invoice, lookup_invoice)

## Description
- Wire processor into settings/factory flow
- Add `paymentsProcessors.alby` config:                                                                                                                                
  - `invoiceExpirySeconds`
  - `replyTimeoutMs`
- Accept both NWC URI schemes for compatibility:
    - `nostr+walletconnect://...`
    - `nostrwalletconnect://...`
- Use `make_invoice` for invoice generation.
- `lookup_invoice` for polling/status reconciliation.
- State mapping to relay invoice lifecycle (pending/accepted/settled/failed/expired -> internal status).
- Configurable reply-timeout enforcement and expiry propagation.
- Update docs and defaults (`README.md`, `CONFIGURATION.md`, `.env.example`, `default-settings.yaml`).

## Related Issue
#323

## How Has This Been Tested?
  - Added processor unit tests for:
    - response/state mapping
    - timeout enforcement from config
    - expiry parameter propagation/mapping
    - timer cleanup on successful fast responses
  - Added factory tests for:
    - required env/config validation
    - legacy URI compatibility

## Screenshots:
image
image

## Types of changes
- [ ] Non-functional change (docs, style, minor refactor)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my code changes.
- [x] I added a changeset, or this is docs-only and I added an empty changeset.
- [x] All new and existing tests passed.

PR #546, opened 2026-04-19
fix: return HTML for browser Accept headers on root route (#532)

Justxd22:
## Description
  - fix `/` content negotiation to serve NIP-11 only for explicit `application/nostr+json` requests                                                                   
  - prevent typical browser `Accept` headers (e.g. `text/html,...,*/*`) from being treated as NIP-11 requests                                                         
  - reuse the same explicit Accept check in router middleware and `rootRequestHandler`                                                                                
  - add unit coverage for explicit NIP-11, browser Accept header, and `q=0`                                                                                           
  - add integration coverage for browser-style Accept header returning HTML 

## Related Issue
#532 

## How Has This Been Tested?
  - `npm run build:check`                                                                                                                                             
  - `npm run test:unit`   

## Types of changes

## Checklist:


Justxd22:
Ci should pass once #539 is merged

cameri:
Did you test the fix manually? Also does it work when the relay URL is in a path other than /?

Justxd22:
@cameri confirmed it breaks under subpath! so We need to edit the html templates remove hardcoded root url and inject `settings.info.relay_url` to get the sub path

Justxd22:
@cameri ready !

Justxd22:
@cameri 

- HTML links, invoice form actions, and invoice status polling now use a rendered `path_prefix`.
- path_prefix is derived from `settings.info.relay_url` by default.
- `X-Forwarded-Prefix` is only honored when the immediate proxy IP is configured in `network.trustedProxies`.
- Added regression tests for trusted and untrusted forwarded prefix behavior.


 ### i suggest Changing default `network.trustedProxies` to [] so forwarded headers are opt-in, not trusted by default.

cameri (approved):

PR #565, opened 2026-04-22
chore(ci): skip expensive jobs on non-source changes

cameri:
## Summary

- Adds a `changes` job using `dorny/paths-filter@v3` to detect whether source-affecting files changed (`src/`, `test/`, `package.json`, `tsconfig*.json`, `biome.json`, `.knip.json`, Dockerfiles, `.nvmrc`, and the workflow file itself)
- Gates `lint`, `build-check`, and both test jobs behind `needs.changes.outputs.src == 'true'` so they skip on docs/config-only PRs
- Fixes `commit-lint` false positives on squash-merges by restricting it to `pull_request` and `workflow_dispatch` only
- Guards `post-tests` Coveralls `parallel-finished` signal to only fire when at least one test job ran (prevents spurious 0% coverage reports)
- Bumps `actions/checkout` and `actions/setup-node` from `@v3` → `@v4`

## Why job-level filter instead of trigger-level `paths:`

GitHub only satisfies required status checks when a job actually runs. A trigger-level `paths:` filter skips the entire workflow, leaving required checks unsatisfied and blocking PRs. The job-level approach keeps the workflow running but skips expensive jobs internally.

## Test plan

- [ ] Open a PR touching only a `.md` file → `lint`, `build-check`, `test-units-and-cover`, `test-integrations-and-cover` show as **Skipped** (green) in Checks
- [ ] Open a PR touching a file in `src/` → all jobs run normally
- [ ] Push directly to `main` → `commit-lint` is skipped, `changesets.yml` runs as before
- [ ] Trigger `workflow_dispatch` → all jobs run (path-filter returns `true` with no diff)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cameri:
@copilot address all the inline comments/reviews

PR #584, opened 2026-04-26
fix: use timing-safe HMAC comparison in Nodeless callback controller

### LINKED ISSUE #581: Security: Nodeless callback uses `!==` for HMAC check and logs expected signature on mismatch
**Describe the bug**

There are two issues in controller callbacks: 

1. The HMAC signature check uses plain `!==` string comparison, which can short-circuit on the first mismatched character and leak timing information. The OpenNode callback already handles this correctly with `crypto.timingSafeEqual`.

2. On signature mismatch, the error log includes the server-computed valid HMAC (`expected`). Since HMAC is deterministic, anyone with log access could replay the same request body with the leaked signature and pass the check.

OpenNode uses `timingSafeEqual`, validates input format, and doesn't log the expected value.

**Expected behavior**

The Nodeless callback should follow the same pattern as OpenNode:
- Use `crypto.timingSafeEqual` with `Buffer` comparison
- Validate the signature header format before comparing
- Only log `actual` (what the caller sent), not `expected`

Anshumancanrock:
## Description

The Nodeless callback controller was using a plain `!==` string comparison to verify webhook HMAC signatures. This PR fixes it to use `crypto.timingSafeEqual` with proper Buffer comparison, matching how the OpenNode callback already handles it.

Also adds a guard for when `NODELESS_WEBHOOK_SECRET` is not set (previously this would throw a `TypeError` from `createHmac`)

## Related Issue

Fixes #581

## Motivation and Context

The OpenNode callback controller uses `timingSafeEqual` for its HMAC check, but the Nodeless controller was using `!==`. This is inconsistent and goes against the standard practice of using constant-time comparison for secret values.

Also, if `NODELESS_WEBHOOK_SECRET` was unset (e.g. on a relay using a different payment processor), any POST to `/callbacks/nodeless` would crash with a `TypeError` instead of returning a clean error response and removes the valid HMAC value from the signature mismatch log (the old code logged `{ expected, actual }`, leaking the correct signature).

## How Has This Been Tested?

- All 1051 existing unit tests pass
- Added 3 new test cases:
  - `returns 403 when callback signature has wrong length`
  - `returns 403 when callback signature is a valid-length hex string but does not match`
  - `returns 500 when NODELESS_WEBHOOK_SECRET is not configured`
- Biome lint passes (227 files, no issues)
- TypeScript build check passes (`tsc --noEmit`)

N/A

## Types of changes

## Checklist:


Anshumancanrock:
hii @cameri , could you please review this PR? Thanks !

Anshumancanrock:
@copilot please review !

src/controllers/callbacks/nodeless-callback-controller.ts
```diff
@@ -30,20 +41,32 @@ export class NodelessCallbackController implements IController {
...
+    const expectedBuf = hmacSha256(webhookSecret, (request as any).rawBody)
+    const actualHex = request.headers['nodeless-signature']
+    const expectedHexLength = expectedBuf.length * 2
 
-    if (expected !== actual) {
-      logger.error('nodeless callback request rejected: signature mismatch:', { expected, actual })
+    if (
+      typeof actualHex !== 'string' ||
+      actualHex.length !== expectedHexLength ||
+      !/^[0-9a-f]+$/i.test(actualHex)
```
cameri: Should we use zod to validate the incoming request? I find much of this section could have been a zod validation, including the timing safe equal, with much less code.

src/controllers/callbacks/nodeless-callback-controller.ts
```diff
@@ -20,6 +22,15 @@ export class NodelessCallbackController implements IController {
     logger('callback request headers: %o', request.headers)
     logger('callback request body: %O', request.body)
 
+    const settings = createSettings()
+    const paymentProcessor = settings.payments?.processor
+
+    if (paymentProcessor !== 'nodeless') {
+      logger('denied request to /callbacks/nodeless which is not the current payment processor')
+      response.status(403).send('Forbidden')
```
cameri: This should probably just be a Bad Request, or even better if we don't even register this route when this payment method isn't enabled it will return a 404 automatically.

src/controllers/callbacks/nodeless-callback-controller.ts
Anshumancanrock: Agreed ! Switched to a Zod schema for the format check. Kept `timingSafeEqual` separate since it needs the computed HMAC at runtime.

src/controllers/callbacks/nodeless-callback-controller.ts
Anshumancanrock: Thanks for the feedback,  I initially mirrored the pattern from the OpenNode controller which also returns 403 for this case, but you're right that not registering the route at all is cleaner. Moved the check to the router level so it 404s automatically when nodeless isn't the active processor.

Anshumancanrock:
hii @cameri , Should I apply the same conditional route registration to the other payment processors (OpenNode, LNbits, Zebedee) ? They still have the in-controller check.

cameri:
@Anshumancanrock yes but in a separate ticket

cameri (approved):

PR #586, opened 2026-04-27
feat: support geohash prefix matching for #g filters

### LINKED ISSUE #265: [REQUEST] search geohash tags by prefix
A typical geohash filter will look for events in a given bounding box, which can be described by a list of geohash prefixes.
Following the [discussion here](https://github.com/nostr-protocol/nips/issues/278#issuecomment-1468892546), I think it would be useful if specifying a "*" at the end of a geohash will mean a search for all events with a geohash tag that contains the requested prefix.

I think the change should be somewhere [here](https://github.com/Cameri/nostream/blob/f23740073f6b59b230157528376a89ca44af3f27/src/repositories/event-repository.ts#L142-L148), but I'm not familiar with Knex. If `filterName[1]` is "g" and `criterion.slice(-1)` is "*", then instead of checking that `event_tags` contain a pair of `[filterName[1], criterion]`, we should check that `event_tags` contain a pair with a first element of "g" and a second element that **starts with** `criterion.slice(0, -1)`.

A possible alternative is to assume that every geohash filter that has less than 9 characters is actually a "prefix filter", even if it doesn't end with "*".

Another crazy option is to index every geohash value with all of its prefixes (i.e. for and event containing `["g","dqcjqcpew"]`, also add to the database `["g","dqcjqcpe"],["g","dqcjqcp"],["g","dqcjqc"],["g","dqcjq"],["g","dqcj"],["g","dqc"],["g","dq"],["g","d"]` - this is crazy but will make the queries simpler.

Thanks!

kanishka0411:
## Description
Adds support for geohash prefix matching for `#g` filters.
A query like `{ "#g": ["u4pruyd*"] }` now matches events with geohash tags starting with `u4pruyd`, for example `["g", "u4pruydqqvj"]`.
Exact `#g` filters without `*` still behave the same as before, so `u4pruyd` does not match `u4pruydqqvj`.

## Related Issue
Fixes #265 

## Motivation and Context
Geohash searches usually work with prefixes, especially when a client is looking for events inside a map area.
Before this, Nostream treated `#g` like any other generic tag filter and only did exact matching. That meant the wildcard form requested in #265, like `u4pruyd*`, did not match longer geohash values.
This updates both paths that need the behavior:
- stored event queries use `LIKE 'prefix%'`
- live/in-memory subscription matching uses `startsWith`

So historical results and live events stay consistent.

## How Has This Been Tested?
Added unit tests for:

- wildcard geohash prefix matching
- exact `#g` matching without wildcard
- generated SQL (`LIKE 'prefix%'` vs `=`)
- in-memory event matching

All passing locally.

## Types of changes

## Checklist:

### DIFF
--- a/src/utils/event.ts
+++ b/src/utils/event.ts
@@ -40,6 +40,18 @@ export const isEventMatchingFilter =
   (filter: SubscriptionFilter) =>
   (event: Event): boolean => {
     const startsWith = (input: string) => (prefix: string) => input.startsWith(prefix)
+    const isMatchingGenericTagCriterion = (key: string, criterion: string) => (tag: Tag): boolean => {
+      const [, tagName] = key
+      if (tag[0] !== tagName) {
+        return false
+      }
+
+      if (key === '#g' && criterion.endsWith('*')) {
+        return tag[1].startsWith(criterion.slice(0, -1))
+      }
+
+      return tag[1] === criterion
+    }
 
     // NIP-01: Basic protocol flow description
 
@@ -84,7 +96,7 @@ export const isEventMatchingFilter =
       Object.entries(filter)
         .filter(([key, criteria]) => isGenericTagQuery(key) && Array.isArray(criteria))
         .some(([key, criteria]) => {
-          return !event.tags.some((tag) => tag[0] === key[1] && criteria.includes(tag[1]))
+          return !event.tags.some((tag) => criteria.some((criterion) => isMatchingGenericTagCriterion(key, criterion)(tag)))
         })
     ) {
       return false
--- a/src/utils/event.ts
+++ b/src/utils/event.ts
@@ -8,7 +8,7 @@ import { deriveFromSecret } from './secret'
 import { EventKindsRange } from '../@types/settings'
 import { fromBuffer } from './transform'
 import { getLeadingZeroBits } from './proof-of-work'
-import { isGenericTagQuery } from './filter'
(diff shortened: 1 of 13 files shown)


src/utils/event.ts
```diff
@@ -40,6 +40,18 @@ export const isEventMatchingFilter =
   (filter: SubscriptionFilter) =>
   (event: Event): boolean => {
     const startsWith = (input: string) => (prefix: string) => input.startsWith(prefix)
+    const isMatchingGenericTagCriterion = (key: string, criterion: string) => (tag: Tag): boolean => {
+      const [, tagName] = key
+      if (tag[0] !== tagName) {
+        return false
+      }
+
+      if (key === '#g' && criterion.endsWith('*')) {
```
cameri: The '#g' should likely be a named constant or encapsulated in a function. I think we keep an enum with those.

See EventTags or isGenericTagQuery.

We should also update the zod schemas to correctly validate the g and #g tags in events and filters so only valid geohashes are passed and the filter allows an optional single * wildcard at the end.

kanishka0411: Both done- ```EventTags.Geohash``` added, ```geohash``` helpers moved into ```src/utils/filter.ts``` so no raw '#g' left. Schemas validate g tags and #g filter values, with the filter allowing one optional trailing *.

cameri (approved):

PR #591, opened 2026-04-30
fix: add husky install fallback

### LINKED ISSUE #328: [BUG] sh: husky: not found
running `./script/start` yields this error: `sh: husky: not found`

Priyanshubhartistm:
## Description
                                                                                                                        
  I fixed the Husky install path that was causing sh: husky: not found.                                                 
                                                                                                                        
  The old prepare script ran husky install  exit 0. That avoided failing the install, but it still printed the shell  
  error first when Husky was not installed.                                                                             
                                                                                                                        
  I moved that logic into .husky/install.mjs so Husky is skipped cleanly in production, CI, HUSKY=0, and installs where 
  the Husky package is not available.                                                                                   
                                                                                                                        
  I also updated the Dockerfiles so .husky/install.mjs is copied before npm install runs.                               
                                                                                                                        
  ## Related Issue                                                                                                      
                                                                                                                        
  Fixes #328                                                                                                            
                                                                                                                        
  ## Motivation and Context                                                                                             
                                                                                                                        
  Husky is only needed for local git hooks. It should not make production, CI, or Docker installs look broken when dev  
  dependencies are missing.                                                                                             
                                                                                                                        
  This keeps local hook setup working, but avoids the noisy sh: husky: not found message in environments where Husky    
  should be skipped.                                                                                                    
                                                                                                                        
  ## How Has This Been Tested?                                                                                          
                                                                                                                        
  I verified the script syntax:                                                                                         
                                                                                                                        
  node --check .husky/install.mjs                                                                                       
                                                                                                                        
  I verified the skip paths:                                                                                            
                                                                                                                        
  NODE_ENV=production npm run prepare --silent                                                                          
  CI=true npm run prepare --silent                                                                                      
  HUSKY=0 npm run prepare --silent
  npm_config_omit=dev npm run prepare --silent                                                                          
                                                                                                                        
  All commands completed successfully without printing the Husky error.                                                 
                                                                                                                        
  I also checked the old behavior still reproduces the original issue:                                                  
                                                                                                                        
  sh -c 'husky install  exit 0'                                                                                       
                                                                                                                        
  It prints:                                                                                                            
                                                                                                                        
  sh: 1: husky: not found                                                                                               
                                                                                                                        
<img width="1448" height="669" alt="image" src="https://github.com/user-attachments/assets/5977f0c1-83af-41d8-931f-c1d737e9776a" />

                                                                                                         
                                                                                                                        
  ## Types of changes                                                                                                   
                                                                                                                        
  ## Checklist:                                                                                                         
                                                                                                                        
  - [x] I have read the CONTRIBUTING document.                                                                          

### DIFF
--- a/.changeset/husky-fallback.md
+++ b/.changeset/husky-fallback.md
@@ -1,0 +1,5 @@
+---
+"nostream": patch
+---
+
+fix: add husky install fallback for non-dev environments
--- a/.husky/install.mjs
+++ b/.husky/install.mjs
@@ -1,0 +1,18 @@
+import { existsSync } from 'node:fs';
+import { execSync } from 'node:child_process';
+
+// Skip Husky installation in environments where hooks are not needed
+if (
+  process.env.NODE_ENV === 'production' ||
+  process.env.CI === 'true' ||
+  process.env.HUSKY === '0' ||
+  !existsSync('.git')
+) {
+  process.exit(0);
+}
+
+try {
+  execSync('npx husky install', { stdio: 'ignore' });
+} catch {
+  process.exit(0);
+}
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "docker:cover:integration": "pnpm run docker:integration:run pnpm exec nyc --report-dir .coverage/integration pnpm run test:integration -- -p cover",
     "postdocker:integration:run": "docker compose -f ./test/integration/docker-compose.yml down",
     "prepack": "pnpm run build",
-    "prepare": "husky install || exit 0",
+    "prepare": "test -f .husky/install.mjs && node .husky/install.mjs || true",
     "changeset:version": "changeset version && pnpm install --lockfile-only",
     "changeset:publish": "changeset publish"
   },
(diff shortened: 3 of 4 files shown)


Dockerfile.railwayapp
```diff
@@ -21,6 +21,7 @@ ENV DB_USER=$PGUSER
 ENV DB_PASSWORD=$PGPASSWORD
 
 COPY ["package.json", "pnpm-lock.yaml","knexfile.js","./"]
+COPY .husky/install.mjs ./.husky/install.mjs
```
cameri: I don't think this is necessary

Dockerfile
```diff
@@ -7,6 +7,7 @@ ARG PNPM_VERSION
 WORKDIR /build
 
 COPY ["package.json", "pnpm-lock.yaml", "./"]
+COPY .husky/install.mjs ./.husky/install.mjs
```
cameri: We don't need husky in the Dockerfile, husky is just for development purposes. Is it causing the Dockerfile build to fail?

Priyanshubhartistm: Removed it,  and yes , COPY was added because the prepare script was calling `install.mjs` during the Docker build which failed without .husky/ in the context. Fixed install.mjs to skip gracefully in non-dev environments instead, so the COPY isn't needed anymore.

Priyanshubhartistm:
@cameri The PR is up for review. Please take a look when you are free.

cameri (approved):

PR #593, opened 2026-05-01
fix: prevent rate limiter bypass via Lua atomic script 

### LINKED ISSUE #467: [BUG]  Sliding-window limiter bypass via non-unique Redis sorted-set members
**Describe the bug**
There is a rate-limiter bypass vulnerability in the `SlidingWindowRateLimiter`. Attackers can easily bypass the configured rate limits by sending a burst of concurrent requests in the exact same millisecond. 

This happens due to two intertwined bugs:
1. **TOCTOU Race Condition:** The `hit` method uses `Promise.all` to concurrently read (`getRangeFromSortedSet`) and write (`addToSortedSet`) to Redis. When hundreds of requests arrive at the exact same millisecond, the Node.js event loop executes the "read" operation for all of them before any "write" operations finish. Redis answers `0` to all initial requests, allowing the entire burst to bypass the limit.
2. **Redis Key Collision (Data Loss):** The rate limiter uses the format `timestamp:step` (e.g., `1775990916270:1`) for Redis sorted set members. Because Redis sorted sets require unique members, when hundreds of requests arrive on the exact same millisecond, they overwrite each other. The server effectively "loses" the count of the vast majority of the burst requests.

**To Reproduce**
1. In `resources/default-settings.yaml` (or `.nostr/settings.yaml`), set a strict rate limit for testing (e.g., `message.rateLimits` to 5 per minute) and ensure `ipWhitelist` is empty.
2. Start the nostream server using Docker (`./scripts/start`).
3. Create a Node.js script using the `ws` package that opens 1000 connections and sends a dummy `REQ` payload simultaneously using `Promise.all()`.
4. Run the script and count the `["NOTICE", "rate limited"]` responses versus accepted responses.
5. See error: The server accepts significantly more requests than the configured limit (e.g., accepting ~60 requests instead of 5).

**Expected behavior**
The server should strictly enforce the rate limit and prevent race conditions. If the limit is 5 per minute, exactly 5 requests should be accepted, and the remaining 995 should accurately receive a "rate-limited" rejection, even if they arrive in the exact same millisecond. No data loss should occur in Redis.

 - OS: Linux
 - Platform: Docker
 - Version: Latest (master)

**Logs**
**Proof of Concept Script Output:**
```text
Connecting to ws://localhost:8008...
1000 connections established. Preparing concurrent payload...
Firing parallel requests in the exact same millisecond...

--- Test Results ---
Total Requests Sent: 1000
Accepted: 59
Rate Limited (Rejected): 941
--------------------
⚠️ BYPASS SUCCESSFUL: More requests were accepted than the configured rate limit (5) allowed.
```
**Additional Context**
**Proposed Fix:**
Two changes are required in `sliding-window-rate-limiter.ts` : 

1. Use a UUID to guarantee uniqueness for every hit in Redis to prevent data loss:
```
import { randomUUID } from 'crypto'
const member = `${now}:${step}:${randomUUID()}`
````
2. Break apart the `Promise.all` in the `hit` method and `await` the Redis operations sequentially so the read accurately reflects the prior write :
```
await this.cache.addToSortedSet(key, member, now.toString())
// ...
const count = await this.cache.getRangeFromSortedSet(key, windowStart)
```

YashIIT0909:
## Description
This PR resolves a critical rate-limiter bypass vulnerability in the [SlidingWindowRateLimiter](cci:2://file:///home/yash-agarwal/Desktop/SOB/nostream/src/utils/sliding-window-rate-limiter.ts:6:0-66:1) by migrating the sliding-window operations from Node.js `Promise.all` logic into an atomic Lua script executed natively on the Redis client. 

It accomplishes three crucial fixes:
1. Eliminates a severe TOCTOU (Time-of-Check to Time-of-Use) race condition caused by interleaved reads and writes during heavy concurrent connection bursts.
2. Dynamically prevents Redis Key Collisions by implementing an entirely self-contained incremental counter loop `while redis.call('ZSCORE', key, member)`, entirely avoiding data loss for exact-millisecond bursts without relying on heavy external UUID libraries for uniqueness.
3. Patches a dormant expiration calculation bug by switching from the seconds-based `EXPIRE` command to the precisely accurate milliseconds-based `PEXPIRE` command.

## Related Issue
Fixes #467

## Motivation and Context
Previously, when hundreds of WebSocket connections or REST payloads arrived in the exact same millisecond, Redis answered `0` to the initial grouped read checks before any writing actually completed due to Node's asynchronous event loop. This allowed massive concurrent bursts to completely bypass the configured metrics. Furthermore, because concurrent writes were uniquely formatted strictly as `timestamp:step`, they overwrote each other as exact-duplicate members in the `ZSET`, effectively erasing accurate metrics from the cache. Utilizing an atomic [eval](cci:1://file:///home/yash-agarwal/Desktop/SOB/nostream/src/@types/adapters.ts:27:2-27:72) wrapper enforces exact, bottlenecked throughput serialization in Redis, guaranteeing the metrics are impenetrable to race-condition exploitation.

## How Has This Been Tested?
1. **Unit Testing:** Updated and successfully passed all existing Mocha unit tests for the [SlidingWindowRateLimiter](cci:2://file:///home/yash-agarwal/Desktop/SOB/nostream/src/utils/sliding-window-rate-limiter.ts:6:0-66:1) and its [eval](cci:1://file:///home/yash-agarwal/Desktop/SOB/nostream/src/@types/adapters.ts:27:2-27:72) implementations via [ICacheAdapter](cci:2://file:///home/yash-agarwal/Desktop/SOB/nostream/src/@types/adapters.ts:19:0-28:1) stubs (`npm run test:unit`).
2. **Security Load Testing:** Deployed a strict `5 / minute` environment threshold locally via Docker Compose.
3. Fired `1,000` concurrent payloads inside the exact same millisecond using a WebSocket parallel `Promise.all()` PoC script.
4. **Result:** Verified that the server perfectly filters and drops `995` requests returning `["NOTICE", "rate limited"]`, while safely accepting exactly `5` requests, proving maximum robustness against intentional bypass bursts.

image

The screenshot is of a checking script that is being ran against a limit of 5.

## Types of changes

## Checklist:


cameri:
@YashIIT0909 The lua script isn't actually being tested, right? How do we know for sure this change is working as intended?

YashIIT0909:
@cameri, as in the attached screenshot, i had created a script that tests it against the limit, and I had modified the limit locally to 5/min and after running the script, and if i ran it again within the same minute, the accepted requests were 0 after that.

cameri (approved):

PR #596, opened 2026-05-02
fix: docker availability precheck

### LINKED ISSUE #590: Discuss alternatives for getRelayUptimeSeconds error handling
## Context

In #585 , `getRelayUptimeSeconds` was patched to wrap 
`runCommandWithOutput` in a try/catch to handle the case where 
Docker is not installed. When Docker is missing, `spawn` emits 
an ENOENT error instead of returning a non-zero exit code, which 
was causing `info --json` to crash and break the pre-commit hook.

## Problem

The current fix works but is a short-term solution. We should 
discuss better long-term alternatives.

## Potential Alternatives

- Pre-check if Docker is installed before invoking the command
- TBD

## Goal

Agree on a consistent approach so similar issues don't arise in 
other command wrappers.

CKodidela:
## Description
  Replaces the per-call-site `try/catch` in `getRelayUptimeSeconds` with a normalized result type on `runCommandWithOutput`. The
  function now always resolves with a `CommandResult` discriminated union instead of rejecting on spawn errors, giving every caller
   a consistent contract.

  ## Related Issue
  Closes #590

  ## Motivation and Context
  `spawn` emits an `ENOENT` error (not a non-zero exit code) when the target command is not installed. Before this change,
  `runCommandWithOutput` forwarded that rejection to callers, which caused `info --json` to crash when Docker was absent. The
  short-term fix was a `try/catch` at the call site, but the same class of bug could recur in any other wrapper.

  The fix moves the concern into the shared runner so all callers benefit automatically.

  ## How Has This Been Tested?
  - Added a regression test (`outputs valid JSON when docker is not installed (ENOENT)`) that stubs `runCommandWithOutput` to
  return `{ ok: false, reason: 'not-found' }` and asserts `info --json` still exits 0 with valid JSON.
  - Updated all existing stubs in `info.spec.ts` and `update.spec.ts` to use the new `CommandResult` shape.
  - All 67 unit tests pass (`npm run test:cli`).

  ## Types of changes

  ## Checklist:


CKodidela:
@cameri up for review

cameri (approved):

CKodidela:
will resolve this soon

cameri (approved):

PR #597, opened 2026-05-02
refactor: guard callback routes with reusable middleware that checks the payment processor

Anshumancanrock:
## Description
The `/callbacks/opennode`, `/callbacks/lnbits`, and `/callbacks/zebedee` routes were always registered regardless of which payment processor was active. Each controller then had to check `settings.payments.processor` and return 403 if it didn't match.

This PR moves that check to the route layer so each route is now only registered when its processor is the configured one, the same way `/callbacks/nodeless` works after #584.

## Related Issue
Follows up on #584 (as approved by @cameri).

## Motivation and Context
The in-controller processor check was redundant , if a route isn't registered, it can't be hit. Keeping dead routes open is unnecessary exposure, and the scattered `paymentProcessor !== 'x'` checks were noise that could easily be missed when reading the controller logic.

## How Has This Been Tested?
- Removed the now-redundant processor-check tests from each controller spec
- Updated `test/unit/routes/callbacks.spec.ts` to stub `createSettings` so the OpenNode route registers correctly during the test

N/A

## Types of changes

## Checklist:


cameri (approved):

src/routes/callbacks/index.ts
```diff
@@ -9,24 +9,30 @@ import { withController } from '../../handlers/request-handlers/with-controller-
 
 const router: Router = Router()
 
-const settings = createSettings()
-const processor = settings.payments?.processor
+const requireProcessor = (name: string) =>
+  (_req: Request, res: Response, next: NextFunction) => {
+    const settings = createSettings()
+    if (settings.payments?.processor !== name) {
```
cameri, line 15: should we also check here if payments are enabled, not just which processor is configured?

cameri:
PR title does not quite match the change: callbacks are always registered, but guarded by a reusable middleware that checks the payment processor is configured before allowing the request.
Changeset has same issue.

src/routes/callbacks/index.ts
Anshumancanrock: > should we also check here if payments are enabled, not just which processor is configured?

Hi @cameri , this PR #616  addresses your comment .`requireProcessor` now checks `payments.enabled` before the processor name. Also fixed the OpenNode integration test setup (it was missing callbackBaseURL, so all 4 scenarios were silently returning 500).

PR #600, opened 2026-05-03
fix: maxlimit check to subscription event requests

### LINKED ISSUE #598: [BUG] maxLimit setting is advertised in NIP-11 relay info but never enforced on subscription queries
**Describe the bug**
The relay exposes `limits.client.subscription.maxLimit `(default: 5000) in the NIP-11 relay information document via `root-request-handler.ts:90`, telling clients that at most 5,000 events will be returned per query. However, this value is never actually enforced anywhere in the query path.

**To Reproduce**
1. Start a nostream relay with default settings (`maxLimit: 5000` in `default-settings.yaml`)
2. Connect via WebSocket and send:
```
["REQ", "test-sub", {"kinds": [1], "limit": 999999}]
```
3. Observe that the relay attempts to stream up to 999,999 events from Postgres - the `maxLimit: 5000` setting has no effect

**Expected behavior**
`findByFilters()` should clamp `currentFilter.limit` to the configured `maxLimit` before passing it to the query builder.

**Screenshots**
Image
The request didn't get flagged, and nostream tried to stream events (though locally I don't have 999999 events stored.)

 - OS: Any
 - Platform: Docker / Standalone
 - Version: 2.1.1 (commit d8f62b4)

**Additional context**
- This creates a broken contract: the relay tells clients the limit is 5,000 but accepts and executes arbitrary values
- Impact: resource exhaustion (memory + DB connections), potential denial of service on relays with large event stores

saniddhyaDubey:
Closes #598 

## Screenshots:
All existing test cases are passing
Screenshot 2026-05-03 at 4 45 31 AM

New test case for the check is added in unit test and passing
Screenshot 2026-05-03 at 4 47 12 AM

## Types of changes

## Checklist:


test/unit/handlers/subscribe-message-handler.spec.ts
```diff
@@ -367,6 +367,23 @@ describe('SubscribeMessageHandler', () => {
       )
     })
 
+    it('returns reason if filter limit exceeds max limit', () => {
```
cameri, line 370: Are we are missing a test case for when the limit is not exceeded?

src/handlers/subscribe-message-handler.ts
```diff
@@ -116,6 +116,14 @@ export class SubscribeMessageHandler implements IMessageHandler, IAbortable {
       }
     }
 
+    const maxLimit = subscriptionLimits?.maxLimit ?? 0
+    if (maxLimit > 0) {
+      const hasExcessiveLimit = filters.some((filter) => filter.limit !== undefined && filter.limit > maxLimit)
```
cameri, line 121: filter.limit != undefined can safely be removed, because filter.limit > maxLimit will never be true if filter.limit isn't a number and greater than maxLimit.

saniddhyaDubey: `limit?: number`, this is the definition of limit in SubscriptionFilter, so I just followed that. Do you want me to remove that and use nullish coalescing approach, something like - `(filter.limit ?? 0)`.

test/unit/handlers/subscribe-message-handler.spec.ts
saniddhyaDubey: I see that is already covered in integration test in feature-1, so this was just to test that specific check. Does that make sense or am I missing something ?

src/handlers/subscribe-message-handler.ts
cameri: What I meant was that `filter.limit !== undefined && filter.limit > maxLimit` can just be `filter.limit > maxLimit`, because `undefined > number` will never be true. We can leave as is, it's just an optimization.

cameri (approved):

PR #616, opened 2026-05-08
fix: check payments.enabled in callback route middleware

Anshumancanrock:
## Description

The `requireProcessor` middleware now also checks `settings.payments.enabled` before allowing callback requests through. Previously it only verified the processor name matched, which meant a relay with payments disabled but a processor configured could still accept callbacks.

## Related Issue

Follow-up to #597

## Motivation and Context

A relay operator might have `payments.enabled: false` with a processor still configured (e.g. during setup or after disabling payments). Without this check, callback endpoints would still accept and process requests so the middleware should reject early in that case.

## How Has This Been Tested?

- All 1261 unit tests passing
- Updated route unit test stub to set `enabled: true`
- Updated OpenNode integration test setup to set `enabled: true`

## Types of changes

- [x] Bug fix

### DIFF
--- a/.changeset/callback-check-payments-enabled.md
+++ b/.changeset/callback-check-payments-enabled.md
@@ -1,0 +1,5 @@
+---
+"nostream": patch
+---
+
+fix: check payments.enabled in callback route middleware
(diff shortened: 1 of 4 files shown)


cameri (approved):

cameri:
Thank you @Anshumancanrock !

Issue #533, opened 2026-04-19
test: integration test verify every path for response type


Justxd22:
write new integration test to check if valid response types for different paths

Ferryx349:
@cameri can i work on this?

cameri:
thanks @Ferryx349 !

PR #627, opened 2026-05-25
test(integration): verify response content-types across HTTP routes

Ferryx349:
Fixes - #533

### DIFF
--- a/.changeset/response-types-integration-tests.md
+++ b/.changeset/response-types-integration-tests.md
@@ -1,0 +1,5 @@
+---
+"nostream": patch
+---
+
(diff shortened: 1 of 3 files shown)


cameri:
Some of these tests are calling endpoints that are disabled, so the tests should likely enable the feature before making the HTTP call so you don't just get a 4xx response.

Ferryx349:
> Some of these tests are calling endpoints that are disabled, so the tests should likely enable the feature before making the HTTP call so you don't just get a 4xx response.

Fixed it, please review

cameri (approved):

PR #622, opened 2026-05-22
feat: add NIP-42 client authentication

Anshumancanrock:
## Description

First of two PRs for NIP-42. This one just adds the types, constants and schema.

- added `EventKinds.AUTH` (22242), `EventTags.Challenge` and `EventTags.AuthRelay` to `constants/base.ts`
- added `MessageType.AUTH`, `AuthMessage` and `AuthChallengeMessage` to `@types/messages.ts`
- added `authMessageSchema` to `message-schema.ts` and wired it into the main `messageSchema` union
- added `createAuthChallengeMessage` helper to `utils/messages.ts`

The handler, WebSocket wiring and session state are in the follow-up PR.

## Related Issue

Closes #619

## Motivation and Context

Needed a clean base before touching the adapter. Keeping the type changes separate makes the actual handler PR much easier to review.

## How Has This Been Tested?

Added schema unit tests for the new `authMessageSchema`. All 1265 tests pass, lint and tsc clean.

## Types of changes
- [ ] Non-functional change (docs, style, minor refactor)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my code changes.
- [x] I added a changeset, or this is docs-only and I added an empty changeset.
- [x] All new and existing tests passed.

PR #629, opened 2026-05-25
feat: Add Nip42 handler

### LINKED ISSUE #619: feat: Add NIP-42 client authentication
**Is your feature request related to a problem? Please describe.**

Right now there's no way to know who's connected. `WebSocketAdapter` tracks `clientId`, `clientAddress`, `alive`, and `subscriptions` but nothing about identity. Every connection is anonymous.

This makes it impossible to:
- restrict publishing to known users
- rate limit by pubkey instead of IP (IPs are trivially spoofed behind proxies)
- build anything on top that needs to know *who* is connected (protected events, access tiers, etc.)

**Describe the solution you'd like**

Implement [NIP-42](https://github.com/nostr-protocol/nips/blob/master/42.md) challenge-response auth:

1. Relay sends `["AUTH", <challenge>]` on connect (challenge = `crypto.randomBytes(32).toString('base64url')`)
2. Client replies with `["AUTH", <signed kind 22242 event>]` containing the challenge and relay URL in tags
3. Relay verifies: kind is 22242, timestamp within 10 min, challenge tag matches what we issued, relay tag matches our URL, Schnorr sig is valid
4. On success, store the authenticated pubkey in per-connection session state
5. Reply `["OK", event.id, true/false, "reason"]`

**Additional context**

I'll split this across a few PRs. This is phase 1 of my SOB 2026 project.

Anshumancanrock:
## Description

Second PR for NIP-42 (follows #622). This one wires everything up so the adapter now sends a challenge on connect and there's a handler to verify the client's AUTH response.

- `WebSocketAdapter` generates a random challenge on connect and sends `["AUTH", <challenge>]`
- new `AuthMessageHandler` verifies kind 22242 events (checks id, sig, timestamp, challenge tag, relay domain)
- on success the pubkey gets stored in a per-connection set, cleared on disconnect
- kind 22242 is blocked from the normal EVENT pipeline so it can't hit storage
- `messageHandlerFactory` routes AUTH messages to the new handler
- extended `IWebSocketAdapter` with `getChallenge()`, `getAuthenticatedPubkeys()`, `addAuthenticatedPubkey()`

## Related Issue

Closes #619

## Motivation and Context

With the types from #622 in place, this wires up the actual auth flow so connections can be tied to a pubkey. Sets things up for NIP-70 protected events next.

## How Has This Been Tested?

Added 20 new tests across the handler, adapter and factory specs.

## Types of changes

## Checklist:


Anshumancanrock:
Hi @cameri , This PR is up for review.

test/unit/handlers/auth-message-handler.spec.ts
```diff
@@ -0,0 +1,233 @@
+import { expect } from 'chai'
```
Ferryx349, line 1: Can we add tests for port-mismatch and re-authentication behaviour.

src/@types/adapters.ts
```diff
@@ -15,6 +15,9 @@ export type IWebSocketAdapter = EventEmitter & {
   getClientId(): string
   getClientAddress(): string
   getSubscriptions(): Map<string, SubscriptionFilter[]>
+  getChallenge(): string
+  getAuthenticatedPubkeys(): ReadonlySet<string>
```
Ferryx349, line 19: A websocket connection should represent exactly one authenticated Nostr pubkey. As NIP-42 says "the client authenticates as a pubkey"~singular. And here set-based design allows multiple identities on one connection. Can we store a single pubkey and replace it on re-authentication, like this:-  

```
getAuthenticatedPubkey(): string | undefined 
```

@cameri wdut

src/handlers/auth-message-handler.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.webSocket.emit(
+      WebSocketAdapterEvent.Message,
+      createCommandResult(eventId, success, message),
+    )
+  }
+
+  // NIP-42 says domain-match is sufficient for relay URL comparison
+  private isRelayUrlMatch(clientRelay: string, serverRelay: string): boolean {
+    try {
+      const clientHost = new URL(clientRelay).hostname.toLowerCase()
```
Ferryx349, line 79: Since `URL.hostname` strips the port, so `wss://relay.example.com` and `wss://relay.example.com:8080` will produce the same `.hostname` value i.e `relay.example.com` and the check will pass for both. 
Can we use `URL.host` instead:-

```
const clientHost = new URL(clientRelay).host.toLowerCase()  
const serverHost = new URL(serverRelay).host.toLowerCase() 
```

Ferryx349 (commented):
Overall the changes look great, just few nits:

src/@types/adapters.ts
Anshumancanrock: Actually NIP-42 explicitly allows clients to auth with multiple pubkeys on the same connection. Clients MAY provide signed events from multiple pubkeys in a sequence of AUTH messages. Relays must treat all pubkeys as authenticated accordingly. So the Set here is by design.

src/handlers/auth-message-handler.ts
Ferryx349: Like we can add it as an improvment, though domain name is sufficient..

test/unit/handlers/auth-message-handler.spec.ts
Anshumancanrock: Since domain-only matching is what NIP-42 recommends, a port difference is expected to still pass. So a "port mismatch rejection" test wouldn't be right here. And re-auth is additive, not a replace.

src/handlers/auth-message-handler.ts
Anshumancanrock: hey, `.hostname` is intentional here. Using `.host` would be stricter than what NIP-42 asks for and could reject valid clients behind proxies or on non-standard ports.

src/@types/adapters.ts
Ferryx349: Got it, I was approaching it from a "one authenticated identity per connection" perspective and hadn't realized NIP-42 explicitly permits multiple AUTH events and authenticated pubkeys on the same connection.

That makes the Set-based design much clearer. Appreciate the explanation and the protocol reference.

cameri: I believe NIP-42 should allow multiple clients to authenticate.

Issue #626, opened 2026-05-24
feat: Web of Trust graph service


saniddhyaDubey:
**Is your feature request related to a problem? Please describe.**
Nostream currently has no way to filter events by social trust - any pubkey can publish regardless of whether the relay owner knows them.

**Describe the solution you'd like**
A WotService that builds a 2-hop trust graph from the relay owner's follow list, exposes isTrusted(pubkey) and isReady(), and swaps atomically on refresh. Pure logic, no process wiring.

**Additional context**
Depends on #623 (merged). Worker wiring and admission guard are separate PRs.

PR #642, opened 2026-06-07
feat: wot core service

saniddhyaDubey:
## Description
Introduces WotService - the core Web of Trust graph logic. Builds a 2-hop trust graph rooted at the relay owner's pubkey by fetching kind-3 follow lists from configured seed relays.
Also patches the wot: config block (#623) with the missing seedRelays field, and updates the settings unit tests accordingly.

## Related Issue
Closes (#626).

## How Has This Been Tested?
14 new unit tests in `test/unit/services/wot-service.spec.ts` cover:
 - Initial state (not ready, no pubkeys trusted)
 - `seedPubkey` always trusted regardless of follower count
 - `minimumFollowers` threshold : pubkeys above pass, pubkeys below are rejected
 - Unknown pubkeys return false
 - In-flight build guard — concurrent `buildGraph()` calls are no-ops
 - `building` flag cleared correctly even when the build throws
 - Tag parsing — non-`p` tags and pubkeys shorter than 64 chars ignored
 - `reset()` clears all state and allows a fresh build

Screenshot 2026-06-07 at 1 38 43 PM
Passing all unit test mentioned above

## Types of changes
- [ X ] New feature (non-breaking change which adds functionality)

## Checklist:
- [ X ] My code follows the code style of this project.
- [ X ] I have added tests to cover my code changes.
- [ X ] I added a changeset, or this is docs-only and I added an empty changeset.
- [ X ] All new and existing tests passed.

### DIFF
--- a/src/services/wot-service.ts
+++ b/src/services/wot-service.ts
@@ -1,0 +1,257 @@
+import { RawData, WebSocket } from 'ws'
+import { randomUUID } from 'crypto'
+
+import { createLogger } from '../factories/logger-factory'
+import { IWotService } from '../@types/services'
+import { WoTSettings } from '../@types/settings'
+
+const logger = createLogger('wot-service')
+
+const PHASE1_TIMEOUT_MS = 5_000
+const PHASE2_BATCH_SIZE = 500
+const PHASE2_CONCURRENCY = 5
+const PHASE2_BATCH_TIMEOUT_MS = 30_000
+
+// Kind 3 — contact / follow list
+const KIND_FOLLOW_LIST = 3
+
+/**
+ * Open a WebSocket to `relayUrl`, send a REQ for `filter`, collect all EVENT
+ * payloads, close on EOSE or timeout, resolve with collected events.
+ */
+function fetchEvents(
+  relayUrl: string,
+  filter: Record<string, unknown>,
+  timeoutMs: number,
+): Promise<any[]> {
+  return new Promise((resolve) => {
+    const subId = `wot-${randomUUID().slice(0, 8)}`
+    const events: any[] = []
+    let settled = false
+
+    const finish = () => {
+      if (settled) {
+        return
+      }
+      settled = true
+      clearTimeout(timer)
+      try { ws.close() } catch { /* ignore */ }
+      resolve(events)
+    }
+
+    const timer = setTimeout(finish, timeoutMs)
+
+    let ws: WebSocket
+    try {
+      ws = new WebSocket(relayUrl, { timeout: timeoutMs })
+    } catch (err) {
+      logger.warn('wot-service: could not create WebSocket to %s: %o', relayUrl, err)
+      clearTimeout(timer)
+      resolve(events)
+      return
+    }
+
+    ws.on('open', () => {
+      ws.send(JSON.stringify(['REQ', subId, filter]))
+    })
+
+    ws.on('message', (raw: RawData) => {
+      try {
+        const msg = JSON.parse(raw.toString('utf8'))
+        if (!Array.isArray(msg)) {
+          return
+        }
+
+        if (msg[0] === 'EVENT' && msg[1] === subId && msg[2]) {
+          events.push(msg[2])
+        } else if (msg[0] === 'EOSE' && msg[1] === subId) {
+          finish()
+        }
+      } catch { /* malformed message — ignore */ }
+    })
+
+    ws.on('error', (err) => {
+      logger.warn('wot-service: WebSocket error for %s: %o', relayUrl, err)
+      finish()
+    })
+
+    ws.on('close', finish)
+  })
+}
+
+/**
+ * Fetch from multiple relays in parallel, deduplicate events by id.
+ * Exported so it can be injected / replaced in tests.
+ */
+export async function fetchFromRelays(
+  relayUrls: string[],
+  filter: Record<string, unknown>,
+  timeoutMs: number,
+): Promise<any[]> {
+  const results = await Promise.all(
+    relayUrls.map((url) => fetchEvents(url, filter, timeoutMs))
+  )
+
+  const seen = new Set<string>()
+  const deduped: any[] = []
+
+  for (const batch of results) {
+    for (const event of batch) {
+      if (typeof event.id === 'string' && !seen.has(event.id)) {
+        seen.add(event.id)
+        deduped.push(event)
+      }
+    }
+  }
+
+  return deduped
+}
+
+/**
+ * Run up to `concurrency` async tasks from `items` at a time.
+ */
+async function runConcurrent<T>(
+  items: T[],
+  concurrency: number,
+  task: (item: T) => Promise<void>,
+): Promise<void> {
+  const queue = [...items]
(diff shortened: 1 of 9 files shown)


src/services/wot-service.ts
```diff
@@ -0,0 +1,257 @@
...
+    const events: any[] = []
+    let settled = false
+
+    const finish = () => {
+      if (settled) {
+        return
+      }
+      settled = true
+      clearTimeout(timer)
+      try { ws.close() } catch { /* ignore */ }
```
cameri: does this clean up the events listening on `open`, `message`, `error`, `close`?

src/services/wot-service.ts
```diff
@@ -0,0 +1,257 @@
...
+          events.push(msg[2])
+        } else if (msg[0] === 'EOSE' && msg[1] === subId) {
+          finish()
+        }
+      } catch { /* malformed message — ignore */ }
+    })
+
+    ws.on('error', (err) => {
+      logger.warn('wot-service: WebSocket error for %s: %o', relayUrl, err)
+      finish()
```
cameri: should we pass the `error` and end with a rejection instead of `resolve()`?

src/services/wot-service.ts
```diff
@@ -0,0 +1,257 @@
...
+export async function fetchFromRelays(
+  relayUrls: string[],
+  filter: Record<string, unknown>,
+  timeoutMs: number,
+): Promise<any[]> {
+  const results = await Promise.all(
+    relayUrls.map((url) => fetchEvents(url, filter, timeoutMs))
+  )
+
+  const seen = new Set<string>()
```
cameri: this dedup in memory is unbounded, so for a popular user following many other usrs this can cause Nostream to get OOM-killed

In these scenarios, we are better off using a bloom filter, and use streams to process the information so the amount of memory used is manageable

src/services/wot-service.ts
```diff
@@ -0,0 +1,257 @@
...
+  private booted = false
+  private building = false
+
+  /**
+   * @param fetcher - relay fetch function. Defaults to the real WebSocket
+   *   implementation. Pass a stub in tests.
+   */
+  public constructor(private readonly fetcher: RelayFetcher = fetchFromRelays) {}
+
+  public async buildGraph(settings: WoTSettings): Promise<void> {
```
cameri: the amount of memory used by this function will also be quite significant, would it be a better choice to try and build the graph using PostgreSQL and use queries instead?

src/services/wot-service.ts
saniddhyaDubey: No, not explicitly. Finish is only clearing the timer, trying to close the socket and resolving the events. Here, the `ws.close()` is not immediate, but we can replace it with `ws.terminate()` for immediate shutdown.
And, for the listener, as soon as the function is finished, they are being garbage collected, so I didn't add the manual removing of these listeners, as it would increase the complexity of the code for this specific use-case.

src/services/wot-service.ts
saniddhyaDubey: The only work for fetchEvents is pull events if there are or else return empty. To keep the design simple, I resolved the errors too as the error here is expected. Do you want me to change it to reject ?

src/services/wot-service.ts
saniddhyaDubey: Looking on a more robust stream based approach to keep memory bounded as well.

src/services/wot-service.ts
saniddhyaDubey: One option to bound the in-memory footprint without switching to PostgreSQL: cap the `followerCount` Map at a configurable max size (e.g. 100k entries). Once a pubkey's count reaches minimumFollowers we stop incrementing, and once the map hits the size cap we stop adding new keys entirely. At 100k entries that's ~15 MB worst case, predictable and manageable. The tradeoff is that pubkeys encountered after the cap is hit get silently excluded, but for most relay operators that's acceptable since the most-followed pubkeys in the network will naturally appear first (only considering `followerCount` Map because Bloom filter + streaming is being implemented to make dedup memory bounded).

That said, the PostgreSQL approach is architecturally cleaner, event_tags already has kind-3 data normalized and indexed, so a GROUP BY + COUNT query runs entirely in the DB and returns only the final trusted pubkey list to Node. Zero accumulation on the Node side and exact dedup via DISTINCT.

The one tradeoff with PostgreSQL: it can only trust pubkeys whose kind-3 events have actually been relayed through this nostream instance, whereas the external relay fetch pulls from damus.io, nostr.band etc. regardless of what's been published here.

Happy to go either direction!

src/services/wot-service.ts
cameri: ok we can keep as-is

src/services/wot-service.ts
cameri: > That said, the PostgreSQL approach is architecturally cleaner, event_tags already has kind-3 data normalized and indexed, so a GROUP BY + COUNT query runs entirely in the DB and returns only the final trusted pubkey list to Node. Zero accumulation on the Node side and exact dedup via DISTINCT.

We should go with PostgreSQL (or Redis for caching) so that we don't have potential OOM issues, introduce artificial caps (100k limit), or surprises (silently excluding a pubkey). Imagine a new pubkey who is just followed and falls outside the cap: this user will remain outside, and the operator will think the relay isn't working.

> The one tradeoff with PostgreSQL: it can only trust pubkeys whose kind-3 events have actually been relayed through this nostream instance, whereas the external relay fetch pulls from damus.io, nostr.band etc. regardless of what's been published here.

If the kind-3 events from seed relays are inserted into our database this trade-off goes away. By the way, we have a static mirroring working implementation. Can it be re-used for WoT? It already connects to other relays and fetches events.

src/services/wot-service.ts
```diff
@@ -0,0 +1,241 @@
...
+      const oneHopRows = await (this.eventRepository as any).masterDbClient('event_tags')
+        .select('tag_value')
+        .whereIn(
+          'event_id',
+          (this.eventRepository as any).masterDbClient('events')
+            .select('event_id')
+            .where('event_pubkey', Buffer.from(settings.seedPubkey, 'hex'))
+            .where('event_kind', KIND_FOLLOW_LIST),
+        )
+        .where('tag_name', 'p')
```
cameri, line 170: @saniddhyaDubey this is outside the database layer and needs to be refactored into a repository

src/services/wot-service.ts
```diff
@@ -0,0 +1,241 @@
...
+          (this.eventRepository as any).masterDbClient('events')
+            .select('event_id')
+            .whereIn('event_pubkey', oneHopPubkeys.map((pk) => Buffer.from(pk, 'hex')))
+            .where('event_kind', KIND_FOLLOW_LIST),
+        )
+        .where('tag_name', 'p')
+        .groupBy('tag_value')
+        .havingRaw('COUNT(*) >= ?', [settings.minimumFollowers])
+
+      const trustedSet = new Set<string>(trustedRows.map((r: any) => r.tag_value as string))
```
cameri, line 222: this can be done in SQL using DISTINCT

PR #643, opened 2026-06-08
feat: add NIP-70 protected event detection util

Anshumancanrock:
## Description

nostream doesn't handle `["-"]` tags right now. Per [NIP-70](https://github.com/nostr-protocol/nips/blob/master/70.md), when an event has this tag it means the author wants it to only be accepted from them directly (after auth). Relays should either reject these outright or check NIP-42 auth first.

Just laying the groundwork in this PR:

- add `EventTags.Protected = '-'` to `constants/base.ts`
- add `isProtectedEvent()` helper to `utils/event.ts`
- 6 unit tests covering the tag detection

The handler-side enforcement comes in the next PR.


Anshumancanrock:
hi @cameri , i have opened #644 which adds the rejection logic on top of this, but it needs the util and constant from here to build. Both PRs are ready for review whenever you have time.

cameri (approved):

PR #644, opened 2026-06-15
feat: reject NIP-70 protected events

Anshumancanrock:
## Description

Adds rejection logic for NIP-70 protected events in `EventMessageHandler`.

- Events with a `["-"]` tag get rejected with `auth-required: this event may only be published by its author` (same message the spec uses in its example).
- Kind-6 reposts that embed a protected event in their `content` are also rejected.
- Placed the check after `canAcceptEvent` so content-length limits kick in before we `JSON.parse` repost content.

## Related Issue

Part of NIP-70 support. Follows #643 which added the detection utility.

## Motivation and Context

NIP-70 says relays MUST reject events with `["-"]` by default. Without this, those events just get stored and spread like any other event, which breaks the whole point of the tag. The repost check also covers the case where someone sticks a protected event inside a kind-6 repost body to get around the outer tag check.

## How Has This Been Tested?

10 unit tests added to `event-message-handler.spec.ts` covering:
- Direct protected tag rejection
- Events with no tags / unrelated tags (no false positives)
- Kind-6 repost with embedded protected event
- Kind-6 repost with clean embedded event
- Empty content, invalid JSON content (graceful handling)
- Non-repost kind with JSON content containing `["-"]` (should pass through)
- Precedence: a repost that is itself protected hits `auth-required` before the embed check
- Malformed embedded tags (non-array `tags` field in parsed JSON)

N/A

## Types of changes

## Checklist:


src/handlers/event-message-handler.ts
```diff
@@ -21,6 +21,7 @@ import {
   isEventSignatureValid,
   isExpiredEvent,
   isFileMessageEvent,
+  isProtectedEvent,
   isRequestToVanishEvent,
   isSealEvent,
```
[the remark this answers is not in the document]
Anshumancanrock: `isProtectedEvent` is provided by the base PR #643 (nip70-protected-event-util). This PR is meant to be merged after #643

src/handlers/event-message-handler.ts
```diff
@@ -224,6 +232,25 @@ export class EventMessageHandler implements IMessageHandler {
...
+      try {
+        const embedded = JSON.parse(event.content)
+        if (
+          Array.isArray(embedded?.tags) &&
+          embedded.tags.some((tag: string[]) => Array.isArray(tag) && tag[0] === EventTags.Protected)
+        ) {
+          return 'blocked: reposts must not embed protected events'
+        }
+      } catch (_e) {
+      }
```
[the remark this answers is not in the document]
Anshumancanrock: Committing this suggestion

src/handlers/event-message-handler.ts
```diff
@@ -224,6 +232,26 @@ export class EventMessageHandler implements IMessageHandler {
...
   }
 
+  protected isProtectedEventBlocked(event: Event): string | undefined {
+    if (isProtectedEvent(event)) {
+      return 'auth-required: this event may only be published by its author'
+    }
+
+    if (event.kind === EventKinds.REPOST && event.content.length > 0) {
+      try {
+        const embedded = JSON.parse(event.content)
```
cameri: should we check if `embedded` is a valid event `eventSchema`, `isEventIdValid()` and `isEventSignatureValid()` before we check if it's protected?

src/handlers/event-message-handler.ts
```diff
@@ -224,6 +232,26 @@ export class EventMessageHandler implements IMessageHandler {
...
+        if (
+          Array.isArray(embedded?.tags) &&
+          embedded.tags.some((tag: string[]) => Array.isArray(tag) && tag[0] === EventTags.Protected)
+        ) {
+          return 'blocked: reposts must not embed protected events'
+        }
+      } catch (_e) {
+        // Ignore invalid JSON: repost content is not a valid embedded event
```
cameri: let's log a warning here

Anshumancanrock: Done!

src/handlers/event-message-handler.ts
Anshumancanrock: Yeah, you're absolutely right.

cameri (approved):

PR #650, opened 2026-06-23
feat: add NIP-43 invite code foundation

Anshumancanrock:
## Description

Invite code infrastructure for NIP-43 relay membership.

`claimCode()` uses a single atomic UPDATE with WHERE guards (not expired, uses remaining) so concurrent claims on the same code are safe without transactions.

`claimed_by` has no FK to `users` so users don't exist yet at claim time.

## Related Issue

NIP-43 implementation

## Motivation and Context

Need a data layer for NIP-43 join/leave handlers.

## How Has This Been Tested?

25 unit tests, full suite 1322 passing.

## Types of changes

## Checklist:


migrations/20260624_120000_create_nip43_invite_codes_table.js
```diff
@@ -0,0 +1,28 @@
+exports.up = async function (knex) {
+  await knex.schema.createTable('invite_codes', (table) => {
+    table.string('code', 64).primary()
+    table.binary('created_by').nullable()
+    table.binary('claimed_by').nullable()
+    table.timestamp('expires_at', { useTz: true }).nullable()
+    table.integer('max_uses').notNullable().defaultTo(1)
+    table.integer('use_count').notNullable().defaultTo(0)
```
cameri: I think these two fields could just be one called `remaining_uses`, once it gets to <= 0 the invite can no longer be used.

Anshumancanrock: Makes sense! I've refactored the schema and repo logic to just use `remaining_uses` .To preserve the "unlimited" feature cleanly, it's now nullable.

cameri (approved):

cameri:
Thank you @Anshumancanrock !

Issue #663, opened 2026-07-01
[Admin Console] : Add observability dashboard UI with Grafana embed


Ferryx349:
## Summary

Add the admin observability dashboard at `/admin/`: password login, live KPI strip via Prometheus-backed SSE (`GET /admin/metrics`), and an embedded Grafana overview dashboard.

## Scope

- Admin HTML shell (login, session, logout)
- Static assets at `/admin/assets/`
- SSE snapshot cards: EPS, rejections, connections, CPU, memory, DB/Redis/Prometheus health
- Grafana in `docker-compose` with provisioned Prometheus datasource and `nostream-overview` dashboard
- Embed Grafana in `/admin/` 

## To Verify locally

```bash
docker compose up -d otel-collector prometheus grafana
pnpm dev
# http://127.0.0.1:8008/admin/

cameri:
Thank you @Ferryx349 !

Issue #594, opened 2026-05-02
feat: Admin Console for relay operators


Justxd22:
Running a Nostream relay still requires too much terminal, SQL, log, and configuration-file knowledge, Operators need to monitor relay health, tune `settings.yaml`, inspect traffic, moderate abusive clients, review admission payments, without SSHing into a server or writing database queries.

This is really important for hosted deployments. A user should be able to deploy the latest Nostream image on a provider, open a secure web admin console, configure the relay, monitor it, and operate it day to day with minimal terminal usage.

I propose to Add an admin console for Nostream. This should be treated as an operator platform, not only a metrics dashboard.

The console should provide:

- Secure admin login, eventually using NIP-98 once merged from SoB !
- Live dashboard for relay health, EPS, active connections, subscriptions, and system resources.
- Visual `settings.yaml` editor that validates changes and applies them through the existing hot-reload path without requiring relay restarts.
- Quick moderation actions for pubkeys, IP addresses, event kinds, rate-limit bypass lists, and admission-fee whitelists.
- Event search and inspection without SQL knowledge, starting with structured relay filters and later using NIP-50 when implemented.
- Admission/payment operations views for sats earned, pending invoices, failed callbacks, paid admissions, and payment processor health.
- Webhook management for admission and payment notifications to generic HTTP, Discord, Slack, and Telegram targets.
- Groundwork for relay operations NIP-89 to manage start/stop/restart any stats/logs for deployed services.

###  This will be shipped over 8 PRs by me + combining Sob NIP-50/98/89 PRs, Dashboard PRs, Metrics PR by other contributors 
CC: Cameri approved this final desgin:
<img width="1536" height="1024" alt="Image" src="https://github.com/user-attachments/assets/e5d67195-0fe5-4351-801f-3a1e644d3f80" />

## Phase 1: Admin backend foundation
Authenticate the relay operator before any admin data is exposed.

- Temporary admin password login until NIP-98 Sob lands :)
- Add admin config settings
- Add protected admin router mounted only when enabled.
- Add admin auth middleware with a replaceable auth interface.
- Add `GET /admin/health`.
- Add `GET /admin/session` or equivalent.
- Add unit tests for disabled-by-default behavior and auth.

## Phase 2: Live metrics
Give operators a fast read on whether the relay is healthy
related work: #439 #464 #468
related issues: #144

- Add cluster-aware runtime counters where possible.
- Add lightweight health checks for DB and Redis.
[...]

PR #666, opened 2026-07-06
feat(admin): add observability dashboard UI with Grafana embeds

Ferryx349:
## Summary
- Add admin observability dashboard at `/admin` with live health, throughput, events, and resource cards
- Stream metrics via existing `/admin/metrics` SSE endpoint with stale-data and fault indicators
- Embed Grafana solo panels (events/sec, websocket connections, CPU, memory) with provisioned `nostream-overview` dashboard
- Add logout route, session cookie handling, and CSP/frame-src support for Grafana
- Harden admin health/metrics timeouts and fail-open rate limiting when Redis is unavailable
- Keep admin console opt-in (`admin.enabled: false` by default)

## Related Issue

Fixes :- #663

## How Has This Been Tested?
- Set `admin.enabled: true` & `ADMIN_PASSWORD` in settings/env, sign in at `/admin`
- Verify health cards update live (SYS/PROM/DB/RDS)
- Stop Redis -> Redis shows `[FAULT]`, dashboard stays reachable
- Start `docker compose up -d otel-collector prometheus grafana` -> Grafana panels load
- Run `pnpm test:unit` for admin routes, grafana utils, event handler, and metrics specs

## Video

https://www.youtube.com/watch?v=cRzJ61NDu0o

Note - I have recorded this video before final UI structure, so screens in attached SS and video might differ but functionalities are same.

# Auth
Screenshot 2026-07-06 at 14 36 29

#  Metrices Dashboard
Screenshot 2026-07-06 at 14 38 11

# Graphs 

Screenshot 2026-07-06 at 14 38 24

# Hamburger (Light Mode)

Screenshot 2026-07-06 at 14 38 51

# Grafana dashboard 

Screenshot 2026-07-06 at 14 39 14

## Types of changes

## Checklist:


Justxd22:
Hello, im wondering why Grafana? and not raw tailwind components? i've the code for the original screens at #594 i was waiting for ground work, Good job on the other PRs!

Ferryx349:
@Justxd22 Sure, we can replace the screen with tailwind components. Actually styling was handled with Bootstrap and custom CSS by me, and Grafana was added specifically for the charting layer.

Ferryx349:
Will it work for you to push tailwind code in seperate PR, then we can make changes accordingly in this PR..
@cameri wdyt?

cameri:
@Ferryx349 @Justxd22 let's do the Tailwind components in a separate PR as it was not an original requirement of this work.

cameri (approved):

Ferryx349:
Fixed Conflicts

docker-compose.yml
```diff
@@ -179,6 +179,29 @@ services:
...
+      GF_AUTH_ANONYMOUS_ENABLED: 'true'
+      GF_AUTH_ANONYMOUS_ORG_ROLE: Viewer
+      GF_SECURITY_ALLOW_EMBEDDING: 'true'
+      GF_SERVER_ROOT_URL: ${GRAFANA_URL:-http://127.0.0.1:3000}
+    volumes:
+      - ./grafana/provisioning:/etc/grafana/provisioning:ro
+      - ./grafana/dashboards:/var/lib/grafana/dashboards:ro
+      - grafana-data:/var/lib/grafana
+    ports:
+      - 127.0.0.1:3000:3000
```
cameri: Can we use another port? 3000 is a very common port and will clash in many systems where other services are running.

src/handlers/event-message-handler.ts
```diff
@@ -357,7 +357,14 @@ export class EventMessageHandler implements IMessageHandler {
         continue
       }
 
-      const isRateLimited = await hit({ period, rate, kinds })
+      let isRateLimited = false
+      try {
+        isRateLimited = await hit({ period, rate, kinds })
+      } catch (error) {
+        // Redis outages should not make the relay reject/abort incoming events.
```
cameri: Not sure about this, it means the relay runs unprotected. If a spammer figures out they can kill Redis for example, then they get free rein to spam after that.
On the other hand if Redis is down the relay is unavailable to clients.

I think the least worse option is to fail-closed when Redis is down.

src/utils/admin-health.ts
```diff
@@ -16,9 +17,17 @@ export interface AdminHealthSnapshot {
   redis: AdminDependencyHealth
 }
 
+export const getAdminDependencyPingTimeoutMs = (): number => {
+  const timeoutCandidate = Number(process.env.ADMIN_DEPENDENCY_PING_TIMEOUT_MS)
+  if (!Number.isFinite(timeoutCandidate) || timeoutCandidate < 100) {
+    return 3000
```
cameri: Let's please name naked literal constants

src/utils/admin-health.ts
```diff
@@ -16,9 +17,17 @@ export interface AdminHealthSnapshot {
   redis: AdminDependencyHealth
 }
 
+export const getAdminDependencyPingTimeoutMs = (): number => {
+  const timeoutCandidate = Number(process.env.ADMIN_DEPENDENCY_PING_TIMEOUT_MS)
+  if (!Number.isFinite(timeoutCandidate) || timeoutCandidate < 100) {
```
cameri: Could you elaborate on why we need this <100 check?

cameri: If we keep this check, we should probably log a warning that this is being overrides otherwise the relay operator will be confused as to why they are changing a setting and it's being ignored

Ferryx349: The < 100 guard avoids sub-100ms timeouts that would mostly produce false unhealthy results. I agree we shouldn’t override silently, I’ll add a warning log when the configured value is ignored and we fall back to the default.

src/handlers/event-message-handler.ts
Ferryx349: Agreed. I’ll update the handler so Redis/rate-limiter errors reject events instead of allowing them through, and adjust the unit test to match.

docker-compose.yml
Ferryx349: I'll switch the default Grafana port to 7777 (still configurable via env) so it doesn't clash with other local services.

docs/REDIS.md
```diff
@@ -11,6 +11,8 @@ Nostream needs.
 
 Rate limiting is implemented using a sliding window strategy, which uses Redis sorted sets to track request timestamps. This allows Nostream to accurately enforce rate limits over a rolling time window rather than a fixed one, preventing clients from bursting requests at window boundaries.
 
+When Redis is temporarily unavailable, Nostream keeps serving the relay and admin HTTP endpoints and treats rate-limiter Redis failures as fail-open (events are not blocked solely because Redis is down). Admin health degrades accordingly (`redis.ok=false`), so operators can detect and recover Redis without taking the relay offline.
```
cameri: this is no longer true

Issue #664, opened 2026-07-02
[Admin Console] : Extract shared settings-config module


Ferryx349:
Extract settings read/validate/path logic from `src/cli/utils/config.ts` into a shared runtime module.

Also export guided field schema from the CLI configure TUI for admin UI use.

cameri:
Thank you @Ferryx349 !

PR #672, opened 2026-07-08
feat(admin): extract shared settings-config module for CLI and admin UI

Ferryx349:
## Description

Extracts settings read/validate/path logic from `src/cli/utils/config.ts` into shared runtime modules so the admin settings editor (Phase 3) can reuse the same primitives as the CLI.
Also exports the guided field schema from the configure TUI for admin UI consumption.

## Why
The admin console settings editor needs the same dot-path get/set, validation, and guided field metadata that the CLI already uses. Duplicating that logic would drift over time; this PR establishes a single source of truth ahead of the admin UI work.

## Related Issue

Fixes:-  #664

## Types of changes

## Checklist:


src/cli/commands/config.ts
```diff
@@ -257,6 +252,5 @@ export const runConfigEnvValidate = async (): Promise<number> => {
 }
 
 export const getConfigTopLevelCategories = (): string[] => {
-  const defaults = loadDefaults() as unknown as Record<string, unknown>
-  return Object.keys(defaults)
+  return getTopLevelSettingCategories()
```
cameri: this function just calls another function, let's remove the extra indirection

src/cli/tui/menus/configure.ts
```diff
@@ -5,16 +5,14 @@ import {
...
 import { tuiPrompts } from '../prompts'
 
-const toCategoryLabel = (key: string): string => {
-  return key
-    .split(/[_\-.]/)
-    .filter(Boolean)
-    .map((part) => part.charAt(0).toUpperCase() + part.slice(1))
-    .join(' ')
-}
+const toCategoryLabel = formatSettingCategoryLabel
```
cameri: we should probably rename the original formatSettingCategoryLabel instead of adding an alias

src/utils/settings-guided-schema.ts
```diff
@@ -0,0 +1,121 @@
...
+  value: string
+  label: string
+  settings: GuidedSettingField[]
+}
+
+export const requireNonEmptySettingValue = (value: string): string | undefined => {
+  return value.trim() ? undefined : 'Value is required'
+}
+
+export const requireSafeNonNegativeIntegerSettingValue = (value: string): string | undefined => {
```
cameri: can we use `zod` instead of rolling our own validation?

src/cli/commands/config.ts
Ferryx349: Removed that wrapper; the TUI now uses the shared function directly.

src/cli/tui/menus/configure.ts
Ferryx349: Renamed formatSettingCategoryLabel to toCategoryLabel.

src/utils/settings-guided-schema.ts
Ferryx349: Replaced the custom safe non-negative integer validation logic with a zod.

cameri (approved):

Issue #683, opened 2026-07-13
[BUG] RedisAdapter.hasKey() always returns true - missing await on client.exists()


Priyanshubhartistm:
**Describe the bug**

`hasKey()` in `src/adapters/redis-adapter.ts` wraps the Redis `EXISTS` call in `Boolean()` without `await`ing the promise, meaning it **always returns `true`** (a non-null Promise is truthy).

```ts
// src/adapters/redis-adapter.ts:52 
public async hasKey(key: string): Promise<boolean> {
  await this.connection
  debug('has %s key', key)
  return Boolean(this.client.exists(key))  // ← Promise object, always truthy!
}
```

**Expected behavior**

Should return `true` only when the key actually exists in Redis.

**Impact**

The method is currently commented-out in `event-message-handler.ts` precisely because it doesn't work:
```ts
// const hasKey = await this.cache.hasKey(`${event.pubkey}:is-admitted`)
// TODO: use cache
```
This forces a full Postgres lookup on every event from every authenticated user, causing unnecessary DB load.

**To Reproduce**
1. Enable payments in `.nostr/settings.yaml`
2. Send any event from an admitted pubkey
3. Observe that `hasKey` always returns `true` regardless of cache state

**System:** Linux, Docker

---

cameri:
Nice catch @Priyanshubhartistm

cameri:
Thank you @Priyanshubhartistm!

PR #684, opened 2026-07-13
fix: await Redis EXISTS call in RedisAdapter.hasKey()

Priyanshubhartistm:
## Description
`RedisAdapter.hasKey()` called `this.client.exists(key)` without `await` and wrapped the result in `Boolean()`. Since `client.exists()` returns a `Promise<number>`, `Boolean(promise)` is always `true` (a non-null object is truthy) regardless of whether the key actually exists in Redis. Fixed by awaiting the call and comparing the resolved count to `0`.

Additionally, the existing unit tests for `hasKey` stubbed `client.exists` with sinon's `.returns()` (a plain synchronous value) rather than `.resolves()` (an actual Promise, matching the real Redis client's behavior). This masked the bug entirely — both tests passed against the broken implementation purely because `Boolean(1)` and `Boolean(0)` happened to match the expected true/false values. Updated both stubs to `.resolves()` so they exercise the real async path.

## Related Issue
#683

## Motivation and Context
`hasKey()` always returning `true` means any caller using it as a cache-presence check would incorrectly treat every key as present. The issue notes this previously forced a full Postgres lookup on every event (the call site had to be commented out because the method didn't work). While the codebase has since moved that particular admission-cache check to `getKey()` instead, `hasKey()` remains a public method on `ICacheAdapter` and should behave correctly for any current or future caller.

## How Has This Been Tested?
- Fixed the two existing `hasKey` unit tests to stub `client.exists` with `.resolves()` instead of  `.returns()`, and verified (by temporarily reverting the source fix) that the corrected test now  fails against the old buggy code with `AssertionError: expected true to be false` — proving it  actually catches this regression.
- Ran the full unit suite: `pnpm run test:unit` — 1393 passing.
- Ran the CLI test suite: `pnpm run test:cli` — 73 passing.
- Ran unit coverage: `pnpm run cover:unit` — passes.
- Ran `pnpm lint`, `pnpm exec biome format` (on the touched files), `pnpm check:deps`,
  `pnpm run build:check`, `pnpm run build`, and `pnpm run verify:cli:bu
- Did not run the full Docker integration suite: confirmed `hasKey()` is not called anywhere in
  current production code paths (the admission-cache check in `event-me
  `getKey()` instead), so this change has no reachable integration surface.

N/A — backend-only change, no UI impact.

## Types of changes
- [ ] Breaking change (fix or feature that would cause existing functio

## Checklist:
- [x] I added a changeset, or this is docs-only and I added an empty ch

### DIFF
--- a/.changeset/redis-haskey-missing-await.md
+++ b/.changeset/redis-haskey-missing-await.md
@@ -1,0 +1,5 @@
+---
+"nostream": patch
+---
+
+fix: await Redis EXISTS call in RedisAdapter.hasKey() so it reflects actual key presence instead of always returning true
--- a/src/adapters/redis-adapter.ts
+++ b/src/adapters/redis-adapter.ts
@@ -52,7 +52,7 @@ export class RedisAdapter implements ICacheAdapter {
   public async hasKey(key: string): Promise<boolean> {
     await this.connection
     logger('has %s key', key)
-    return Boolean(this.client.exists(key))
+    return (await this.client.exists(key)) > 0
   }
 
   public async getKey(key: string): Promise<string> {
--- a/test/unit/adapters/redis-adapter.spec.ts
+++ b/test/unit/adapters/redis-adapter.spec.ts
@@ -76,7 +76,7 @@ describe('RedisAdapter', () => {
 
   describe('hasKey', () => {
     it('awaits connection and calls client.exists with the key', async () => {
-      client.exists.returns(1)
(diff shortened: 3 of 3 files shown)


Priyanshubhartistm:
Hii @cameri  , could you please review this PR ? Whenever you're free,I'd really appreciate your feedback.
Thanks

cameri (approved):

PR #702, opened 2026-07-20
feat(nip42): restrict reads of encrypted kinds to authenticated recipients

### LINKED ISSUE #699: [BUG]  NIP-42 auth not enforced on reads (kinds 4, 1059 leak to anyone)
**Describe the bug**

Kind 4 DMs and kind 1059 gift wraps are handed out to anyone who asks for them. We support NIP-42 AUTH but never actually check it on reads, so an unauthenticated client can just `REQ` for these kinds and pull down everyone's encrypted messages, ciphertext plus who-talks-to-who metadata. COUNT and live broadcasts leak the same way.

**To Reproduce**

1. Run the relay (v3.0.0) with some users exchanging kind 4 / 1059 events.
2. Connect without authenticating and send `["REQ", "sub", {"kinds": [4]}]`.
3. You get back every user's DMs, including ones you're neither the author nor a `p` recipient of.
4. `COUNT` and post-subscribe live events behave the same.

**Expected behavior**

Restricted kinds should only go to clients that have authenticated as the event author or a tagged `p` recipient. Needs to hold for REQ, live broadcasts, and COUNT.

COUNT can't be filtered per-event, so any COUNT touching a restricted kind has to be scoped to the caller's own pubkeys via `authors`/`#p`.

If a subscription only asks for restricted kinds and the client isn't authenticated, close it with `auth-required:` so the client knows to auth and retry, instead of silently returning nothing.

Anshumancanrock:
## Description

We already support NIP-42 client authentication (the AUTH handshake , but nothing ever *used* that authentication to gate reads. Any client that is authenticated or not could `REQ` for kind `4` / `1059` and pull down everyone's encrypted DMs and gift wraps, ciphertext plus who-talks-to-who metadata. COUNT and live broadcasts leaked the same way.

This PR adds opt-in auth-based read filtering for a configurable set of "restricted" kinds. When enabled, events of those kinds are only delivered to clients that have authenticated as the event's author or as a pubkey listed in the event's `p` tags. It applies consistently across all three read paths:

- **REQ / stored events** : restricted events are dropped from the stream unless the client is authorized.
- **Live broadcasts** :  same guard in `WebSocketAdapter.onSendEvent`.
- **COUNT** :  counts can't be filtered per event, so any filter touching a restricted kind must be scoped to the caller's own authenticated pubkeys via `authors`/`#p`, otherwise the count is refused.

When an unauthenticated client opens a subscription that *exclusively* targets restricted kinds (it could never return anything), the relay closes it with `auth-required:` reason so the client knows to authenticate and retry, instead of silently returning an empty stream. Mixed filters (e.g. `kinds: [1, 4]`) are still served, the unrestricted events come back and the restricted ones are filtered out per event.

## Related Issue

Closes #699 

## Motivation and Context

A relay that supports NIP-42 but never enforces it on reads offers no real read privacy so a passive observer can harvest every user's encrypted messages and social graph. Enforcing auth on the restricted kinds is what makes NIP-42 useful for private/paid relays and for protecting DM metadata.

## How Has This Been Tested?

- New unit suite `test/unit/utils/nip42.spec.ts` covering the authorization primitives: author match, p-tag recipient match, stranger rejection, kind ranges, mixed vs. fully-restricted filters, COUNT scoping (`authors`/`#p`), and that the auth lookup is *not* consulted when the feature is disabled or the kind is unrestricted (lazy evaluation).

- Added cases to `subscribe-message-handler.spec.ts` (REQ filtering + `auth-required` CLOSED), `count-message-handler.spec.ts` (COUNT scoping / refusal), and `web-socket-adapter.spec.ts` (live broadcast delivered only to the authenticated author/recipient).

- Full relevant suite: **111 passing**. `biome lint` clean on all changed files.

N/A

## Types of changes

## Checklist:


src/utils/nip42.ts
```diff
@@ -0,0 +1,112 @@
...
+    return []
+  }
+
+  return Array.isArray(restrictedReads.kinds) ? restrictedReads.kinds : DEFAULT_RESTRICTED_READ_KINDS
+}
+
+const isKindRestricted = (restrictedKinds: (EventKinds | EventKindsRange)[], kind: number): boolean =>
+  restrictedKinds.some(isEventKindOrRangeMatch({ kind } as Event))
+
+export const isClientAuthorizedToReadEvent = (event: Event, authenticatedPubkeys: ReadonlySet<Pubkey>): boolean => {
```
cameri: This should probably be named: isClientAuthorizedToReadMention()
or canAuthClientReadMention() is a bit shorter.

Since we are not checking the author of the event but who this event is directed to.

src/utils/nip42.ts
```diff
@@ -0,0 +1,112 @@
...
+  const restrictedKinds = getRestrictedReadKinds(settings)
+  if (!restrictedKinds.length) {
+    return false
+  }
+
+  if (!filters.length || !filters.every(isFullyRestrictedFilter(restrictedKinds))) {
+    return false
+  }
+
+  return getAuthenticatedPubkeys().size === 0
```
cameri, line 87: This returns true if the user has not authenticated yet (list of auth pubkeys is empty), is there a guard before it to ensure we can't get here with the user being authenticated?

It's not clear to me what checking the size of authenticated pubkeys does in this context.

src/utils/nip42.ts
Anshumancanrock: Done, renamed it to isClientAuthorizedToReadMention. Kept the author check in there too since the author needs to read their own DMs/gift-wraps.

src/utils/nip42.ts
Anshumancanrock: No guard needed there because an authenticated client is meant to fall through. `createReadAuthorizationGuard` filters restricted events per-event, so they still only get the ones they authored or are p-tagged in. The size === 0 check is just for the unauthenticated case, whose stream would always be empty, so we close it with auth-required instead.

cameri (approved):

PR #708, opened 2026-07-22
fix: reject expiration timestamp 0 and accept far-future timestamps in getEventExpiration()

### LINKED ISSUE #707: [BUG] getEventExpiration() accepts expiration timestamp 0 as valid and rejects valid year 2287+ timestamps
**Describe the bug**

In `src/utils/event.ts`, `getEventExpiration()` validates the NIP-40 `expiration` tag with a `Math.log10()` bounds check:

```ts
if (Number.isSafeInteger(expirationTime) && Math.log10(expirationTime) < 10) {
  return expirationTime
}
```

**To Reproduce**

1. Publish an event with an expiration tag of 0, e.g. ["expiration", "0"].
2. Observe the relay treats it as a validing/ignoring it.
3. Publish an event with an expiration tag > 9999999999, e.g. ["expiration", "10000000000"].
4. Observe the expiration is silently drores.

**Screenshots**

N/A

**System** 

- OS: Linux
- Platform: Docker
- Version: 3.0.0

Logs

N/A — no error is logged; the timestamp ided.

**Additional context**

getEventExpiration() is at src/utils/evenis at src/utils/event.ts:256). Its result feeds addExpirationMetadata() in src/handlers/event-message-handler.ts, so discard-far-future cases affect real event expiry behavior.

Priyanshubhartistm:
## Description
`getEventExpiration()` in `src/utils/event.ts` used `Math.log10(expirationTime) < 10` as a bounds check on the NIP-40 `expiration` tag. This has two bugs:
- `Math.log10(0) === -Infinity`, which is `< 10`, so an expiration of `0`  (Unix epoch) was accepted as valid.
- Any safe-integer timestamp past year 2287 (`> 9,999,999,999`) failed the  `< 10` check and was silently discarded, treated as "no expiration."

Replaced the check with `expirationTime > 0`, which accepts any positive safe integer and rejects non-positive values, with no arbitrary digit-count ceiling.

## Related Issue
Closes #707

## Motivation and Context
An expiration of `0` should never be treated as a valid future expiration, and a relay shouldn't silently ignore legitimate (if far-future) expiration timestamps just because they cross an arbitrary digit-count boundary.

## How Has This Been Tested?
- Added two unit tests in `test/unit/utils/event.spec.ts` under the existing `NIP-40 > getEventExpiration` suite:
  - `expiration: '0'` → `undefined`
  - `expiration: '10000000000'` (year 2287+) → returned as-is
- Ran `pnpm run test:unit` (1395 passing) and `pnpm run test:cli` (73 passing)
- Ran `pnpm run lint`, `pnpm run check:deps`, `pnpm run build`,
  `pnpm run verify:cli:build`, `pnpm run cover:unit` — all pass.

N/A — pure logic fix, no UI/API surface change.

## Types of changes

## Checklist:


src/utils/event.ts
```diff
@@ -253,7 +253,7 @@ export const getEventExpiration = (event: Event): number | undefined => {
 
   const expirationTime = Number(rawExpirationTime)
 
-  if (Number.isSafeInteger(expirationTime) && Math.log10(expirationTime) < 10) {
+  if (Number.isSafeInteger(expirationTime) && expirationTime > 0) {
```
cameri: I think we need all three checks.

The log10 check was so that expiration was not passed in milliseconds. Maybe there's a better way to ensure that?

Priyanshubhartistm: Pushed a fix keeping all three checks  replaced the digit-count check with a fixed ceiling so ms-scale values are still rejected but far-future seconds-based expirations aren't.

phoenix-server (approved):

cameri (approved):

src/utils/event.ts
```diff
@@ -253,7 +258,7 @@ export const getEventExpiration = (event: Event): number | undefined => {
 
   const expirationTime = Number(rawExpirationTime)
 
-  if (Number.isSafeInteger(expirationTime) && Math.log10(expirationTime) < 10) {
```
Anshumancanrock, line 256: Thanks for the PR! The `Math.log10(0)` fix is a genuine improvement.

test/unit/utils/event.spec.ts
```diff
@@ -691,6 +691,26 @@ describe('NIP-40', () => {
...
+      expect(getEventExpiration(event)).to.equal(253402300799)
+    })
+
+    it('returns false if expiration looks like a millisecond-scale timestamp', () => {
+      event.tags = [['expiration', '1700000000000']]
+      expect(getEventExpiration(event)).to.be.undefined
+    })
   })
 
   describe('isExpiredEvent', () => {
```
Anshumancanrock, line 721: There's a problem with the new ceiling though.

`events.expires_at` is a plain PostgreSQL `integer`. The `.unsigned()` in the migration is MySQL specific, and Knex ignores it on PostgreSQL. That means we're capped at `2147483647` (January 2038), so both values your tests accept (`10000000000` and `253402300799`) are too large for the column. In fact, timestamps from 2038 to 2286 already fail on `main` today for the same reason, since the old `log10 < 10` check allowed them through as well.

The `INSERT` then fails, `create()` catches the error and returns `0`, and `DefaultEventStrategy` interprets that as a duplicate. The client ends up getting `["OK", id, true, "duplicate:"]` for an event that was never stored.

On `main`, the same event is at least stored with `expires_at = NULL`. Ignoring the expiration is the bug we're trying to fix, but the note itself is still preserved. With this change, the note is silently dropped instead. So it doesn't quite fix #707; it changes the failure mode from "expiration ignored" to "event silently lost."

The tests only exercise the validation function and never hit the database, so there wasn't really a way for this to be caught by the current test suite. It's an easy thing to miss.

Anshumancanrock (changes requested):

Anshumancanrock:
@Priyanshubhartistm To be clear this isn't on you. 2038–2286 already fails on main for the same reason, the old `log10 < 10` check let those through too.

@cameri i think your ms-scale concern still holds, just at a lower ceiling.

Anshumancanrock:
@Priyanshubhartistm Flip the two far-future tests to expect undefined and add one for `2147483647` and all good.

Actual year-9999 support needs `expires_at` moved to `bigint`. I can open a seperate issue for it.

Priyanshubhartistm:
@Anshumancanrock , thanks for tracing it through to the column type  pushed a fix that caps at 2147483647 to match the actual expires_at int4 column, and flipped/added the tests accordingly. Agreed the year-9999 case is a separate bigint migration; happy to have that tracked as its own issue rather than bundled here

test/unit/utils/event.spec.ts
Priyanshubhartistm: done

cameri (approved):

cameri:
@Anshumancanrock please re-review

Anshumancanrock (approved):
LGTM

cameri:
@Priyanshubhartistm feel free to merge 
@Anshumancanrock thank you!

PR #722, opened 2026-08-08
feat(nip98): add Authorization header verifier

Anshumancanrock:
## Description

First chunk of NIP-98 support. Adds a standalone verifier for `Authorization: Nostr <base64-event>` headers.

- `verifyNip98Auth()` in `src/utils/nip98.ts`
- checks kind `27235`, timestamp skew, exact `u`/`method`, optional body `payload` hash, then id + sig
- oversized headers get rejected early
- also fixed a small `sinon` redeclare in `info.spec.ts` that was breaking the full unit suite

Middleware / admin API wiring comes next.

## Related Issue

Part of NIP-98 admin auth work.

## Motivation and Context

Need a solid crypto check we can drop into middleware later. Keeping it as a pure util makes this easy to review on its own.

## Types of changes

## Checklist:

### DIFF
--- a/src/utils/nip98.ts
+++ b/src/utils/nip98.ts
@@ -1,0 +1,271 @@
+import { createHash, timingSafeEqual } from 'crypto'
+import { z } from 'zod'
+import { Pubkey } from '../@types/base'
+import { Event } from '../@types/event'
+import { EventKinds, EventTags } from '../constants/base'
+import { createdAtSchema, idSchema, kindSchema, pubkeySchema, signatureSchema, tagSchema } from '../schemas/base-schema'
+import { isEventIdValid, isEventSignatureValid } from './event'
+
+// NIP-98 suggests a ~60s window for kind 27235 auth events.
+export const DEFAULT_NIP98_MAX_SKEW_SECONDS = 60
+
+// A signed kind-27235 event is typically ~1–2KB encoded. Cap well above that to
+// reject pathological Authorization headers before JSON.parse / crypto work.
+export const DEFAULT_NIP98_MAX_AUTHORIZATION_HEADER_LENGTH = 8192
+
+const NOSTR_AUTH_SCHEME = /^Nostr$/i
+const LOWER_HEX_64 = /^[0-9a-f]{64}$/
+const BASE64_TOKEN = /^[A-Za-z0-9+/]+={0,2}$/
+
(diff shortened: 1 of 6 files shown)


src/utils/nip98.ts
```diff
@@ -0,0 +1,271 @@
...
+  const scheme = trimmed.slice(0, spaceIndex)
+  const token = trimmed.slice(spaceIndex + 1).trim()
+  if (!NOSTR_AUTH_SCHEME.test(scheme) || token.length === 0) {
+    return fail('invalid authorization scheme')
+  }
+
+  if (!BASE64_TOKEN.test(token) || token.length % 4 !== 0) {
+    return fail('invalid authorization encoding')
+  }
+
```
cameri, line 165: This code could just be a single regular expression that matches the scheme, space, and token in one go.
After it matches, it'd be safe to use .split(' ') and extract the token.

I'd also cap the length from 8 kB to 2 kB: strict rather than lenient. The cost of Auth is that an attacker will happily consume all your resources to deny service to everyone else.

src/utils/nip98.ts
```diff
@@ -0,0 +1,271 @@
...
+  return Buffer.from(token, 'base64').toString('utf8')
+}
+
+const isHexEqual = (left: string, right: string): boolean => {
+  if (left.length !== right.length || !LOWER_HEX_64.test(left) || !LOWER_HEX_64.test(right)) {
+    return false
+  }
```
cameri: This fast path defeats the benefit of using timingSafeEqual below.

Unclear why we need to ensure the strings are hex again, didn't we validate that earlier? Is it possible that we can get here and compare anything that isn't already a hex string?

src/utils/nip98.ts
```diff
@@ -0,0 +1,271 @@
...
+  const token = trimmed.slice(spaceIndex + 1).trim()
+  if (!NOSTR_AUTH_SCHEME.test(scheme) || token.length === 0) {
+    return fail('invalid authorization scheme')
+  }
+
+  if (!BASE64_TOKEN.test(token) || token.length % 4 !== 0) {
+    return fail('invalid authorization encoding')
+  }
+
+  return Buffer.from(token, 'base64').toString('utf8')
```
cameri: We should probably parse the JSON in here rather than letting the caller do it. Only here we know for sure the type is correct, otherwise caller needs to deal with unknown type and cast back to a proper type.

cameri (approved):

PR #730, opened 2026-08-13
feat(admin): accept NIP-98 auth on admin API routes

Anshumancanrock:
### Description

Protected admin API routes now accept `Authorization: Nostr` on top of the existing session cookie.

Off by default. Set `admin.nip98.enabled` and put hex pubkeys in `admin.nip98.allowedPubkeys`. Empty allowlist means nobody gets in that way.

For PATCH/POST we check the header (sig + allowlist) before parsing JSON, so a junk `Nostr` token never hits the 1mb body parser. After the body is read we check the payload hash and store the event id in Redis with `SET NX` (TTL = `maxSkewSeconds`). Same event twice, or Redis down → 401.

GETs don't parse a body. Same session-or-NIP-98 check, including the one-time event id.

Clients still sign `u` against `info.relay_url` (not the request `Host`). Password login / dashboard / cookie session are unchanged; `/login` stays 100kb.

### Related Issue

Follows #722 (verifier), #725 (absolute URL), #726 (Redis NX).

### Motivation and Context

#722 only verifies a header. This is the part that actually lets you call the admin API with a signed event instead of sharing the password.


Anshumancanrock:
I think CodeQL doesn’t recognize our Redis adminRateLimitMiddleware (it only models express-rate-limit etc). These routes still run that limiter before auth, same as the rest of this file.

Ferryx349:
@Anshumancanrock totally agree, its a codeql false positive, it was happening same with me too, had tried a fix in #692.

src/handlers/request-handlers/admin-auth-middleware.ts
```diff
@@ -1,19 +1,187 @@
...
+const adminAuthProvider: IAdminAuthProvider = createAdminAuthProvider()
+
+const METHODS_WITH_BODY = new Set(['POST', 'PUT', 'PATCH', 'DELETE'])
+
+export const isNostrAuthorizationHeader = (authorizationHeader: string | undefined): boolean => {
+  if (typeof authorizationHeader !== 'string') {
+    return false
+  }
+
+  return /^Nostr\s+/i.test(authorizationHeader.trim())
```
cameri: Let's remove the .trim() and change the regex to a full match rather than partial (include the actual token in the match). Not sure why we'd want to allow spaces after the token or before a the Nostr string. Do we strip that later too? I don't think the spec is this loose.

src/handlers/request-handlers/admin-auth-middleware.ts
```diff
@@ -1,19 +1,187 @@
...
+}
+
+const isAllowedNip98Pubkey = (pubkey: string, allowedPubkeys: string[] | undefined): boolean => {
+  if (!Array.isArray(allowedPubkeys) || allowedPubkeys.length === 0) {
+    return false
+  }
+
+  const normalized = pubkey.toLowerCase()
+  return allowedPubkeys.some((allowed) => typeof allowed === 'string' && allowed.toLowerCase() === normalized)
+}
```
cameri:
```suggestion
const isAllowedNip98Pubkey = (pubkey: string, allowedPubkeys: string[] = []): boolean => {
  const normalized = pubkey.toLowerCase()
  return allowedPubkeys.some((allowed) => typeof allowed === 'string' && allowed.toLowerCase() === normalized)
}
```
Would be equivalent.

src/handlers/request-handlers/admin-auth-middleware.ts
```diff
@@ -1,19 +1,187 @@
...
+const resolveBodyForNip98 = (request: AdminRequest): Buffer | undefined | 'missing-raw-body' => {
+  if (request.rawBody !== undefined) {
+    return request.rawBody
+  }
+
+  if (!METHODS_WITH_BODY.has(request.method.toUpperCase())) {
+    return undefined
+  }
+
+  const contentLength = Number(request.headers['content-length'] ?? '0')
```
cameri: can we check this against the actual request body?

src/handlers/request-handlers/admin-auth-middleware.ts
```diff
@@ -1,19 +1,187 @@
...
+  const transferEncoding = Array.isArray(transferEncodingHeader)
+    ? transferEncodingHeader.join(',')
+    : (transferEncodingHeader ?? '')
+  const hasChunkedBody = transferEncoding.toLowerCase().includes('chunked')
+
+  if ((Number.isFinite(contentLength) && contentLength > 0) || hasChunkedBody) {
+    return 'missing-raw-body'
+  }
+
+  return Buffer.alloc(0)
```
cameri: This function `resolveBodyForNip98` is trying doing too much, it returns a Buffer, or a string, or undefined. Will address async.

cameri (approved):

PR #732, opened 2026-08-16
feat: add a CLI to mint NIP-43 invite codes

Anshumancanrock:
### Description

Adds `nostream invite create` so one can mint a NIP-43 claim without writing SQL.

```bash
nostream invite create
nostream invite create --uses 3 --expires-in 86400 --json

# Docker (Postgres isn't published on the host)
docker compose exec nostream node src/cli/index.js invite create
```

Defaults come from `nip43.defaultMaxUses` (1) and `nip43.inviteCodeExpiry` (0 = never). If `info.self` is hex or `npub1…`, it gets stored as `created_by`.

## Related Issue

Follow-up to #676. Join (28934) is already on main; there was still no operator path to create a code.

## Motivation and Context

Kind 28934 can claim a code, but minting one meant inserting into `invite_codes` by hand. That's not usable.


resources/default-settings.yaml
```diff
@@ -73,6 +73,10 @@ nip43:
   # (users who claimed an invite code via a kind 28934 join request) may
   # publish events. Enabling this on an open relay blocks all non-members.
   enabled: false
+  # Seconds until a newly minted invite code expires. 0 means the code never expires.
+  inviteCodeExpiry: 0
```
cameri: Let's set the default to a sane 10 minutes, not infinite. Let's also rename this to `inviteCodeExpirySeconds`

cameri (approved):

resources/default-settings.yaml
Anshumancanrock: Done, renamed to `inviteCodeExpirySeconds` and defaulted it to 600.

src/repositories/invite-code-repository.ts
```diff
@@ -35,18 +33,21 @@ export class InviteCodeRepository implements IInviteCodeRepository {
...
+    const remainingUses = options.remainingUses === undefined ? 1 : options.remainingUses
+    const createdBy = options.createdBy ?? null
+
     logger('create invite code (expires: %s, remainingUses: %s)', expiresAt ?? 'never', remainingUses ?? 'unlimited')
 
     const now = new Date()
     const row: DBInviteCode = {
       code,
-      created_by: null,
+      created_by: createdBy ? toBuffer(createdBy) : null,
```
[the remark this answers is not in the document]
Anshumancanrock: `createdBy` is already validated as 64-char hex in `issueInviteCode` (and the CLI only passes it after parseRelayPubkey). Same `toBuffer` pattern as the other `pubkey` columns.. No change needed here.

cameri (approved):

PR #729, opened 2026-08-12
feat(dvm): trap NIP-90 job request events and record them via job repository

Priyanshubhartistm:
## Description

Adds ingestion for NIP-90 DVM job request events (kind 5000-5999): the relay now recognizes these events, stores and broadcasts them like any other event, and records job state via the `DvmJobRepository` added in #727. No worker dispatch yet  that's a separate follow-up PR.

- Added `EventKinds.DVM_JOB_REQUEST_FIRST`/`DVM_JOB_REQUEST_LAST` (5000/5999) to  `src/constants/base.ts`, and `isDvmJobRequestEvent()` to `src/utils/event.ts`, following the  same range-check pattern as `isEphemeralEvent`/`isParameterizedReplaceableEvent`.
- New `DvmJobRequestEventStrategy` (`src/handlers/event-strategies/dvm-job-request-event-strategy.ts`):
  stores the event via `eventRepository.create()` and broadcasts it (job request events are still regular Nostr events, so clients/DVMs can query them normally), then records job state via `dvmJobRepository.create(id, pubkey, kind)` as a best-effort side effect — a job-repository failure is logged but never causes a valid event to be rejected.
- Wired into `eventStrategyFactory`, checked early alongside the existing NIP-43 checks.
- Threaded `IDvmJobRepository` through the factory chain that constructs the strategy:
  `worker-factory.ts` (constructs `DvmJobRepository`) → `websocket-adapter-factory.ts` →
  `message-handler-factory.ts` → `event-strategy-factory.ts`.
- Removed `src/repositories/dvm-job-repository.ts` from `.knip.json`'s ignore list — it now has a
  real consumer, so the temporary entry added in #727 is no longer needed.

## Related Issue

Fixes #728, part of #639

## Motivation and Context

Kind 5000-5999 currently falls outside every existing event-kind range constant, so DVM jobrequests silently land in  `DefaultEventStrategy` and are treated as plain events — no job state is tracked at all. This PR adds recognition and job-state recording so the persistence layer added in #727 actually gets populated, without yet building the IPC dispatch logic that consumes it (kept separate to avoid a large, hard-to-review PR).

## How Has This Been Tested?

[...]

Issue #639, opened 2026-05-31
Build the IPC Job Queue for NIP-90


Priyanshubhartistm:
## Description

This is the first step toward turning Nostream into a DVM (Data Vending Machine) orchestrator. Before we can do billing, service discovery, we need a way for the relay to actually receive DVM job requests and hand them off to external worker processes.

That's what this issue covers.

## What needs to be done

- Listen for incoming NIP-90 DVM request events (kind `5xxx`)
- Parse the job details (input data, requested compute type, etc.)
- Spawn or communicate with an external process (like a Python script or Node worker) using standard I/O (stdin/stdout) or TCP sockets
- Send the job payload to that process, wait for the result, and relay it back as a NIP-90 result event (kind `6xxx`)
- Queue jobs properly. We need a basic async queue so things don't pile up or get lost

PR #734, opened 2026-08-19
feat(dvm): dispatch pending DVM jobs to worker processes

### LINKED ISSUE #731: Dispatch DVM jobs to worker processes and publish results back
## Description

Follow-up to #639, #727 (job persistence), and #729 (job ingestion). The relay can now recognize NIP-90 job request events and record them as `submitted` jobs, but nothing actually hands them off to a worker process yet.

This issue covers dispatch: pick up a pending job, spawn/communicate with its configured worker, wait for a result (with a timeout), update job state through its lifecycle, and publish the result back as a NIP-90 event.

## What needs to be done

- Pick up `submitted` jobs (via `DvmJobRepository.findPendingJobs()`) and assign them to a worke  (`DvmJobRepository.assignWorker()`, already atomic/race-safe from #727).
- Spawn/communicate with the worker process, extending the existing `spawn()`-based pattern in `src/cli/utils/process.ts` (timeout/kill/error-classification) rather than inventing a second spawn convention.
- Frame messages to/from the worker as newline-delimited JSON over stdin/stdout  no existing convention covers this (process.ts buffers all output into one string, which doesn't work for a long-lived worker handling multiple in-flight jobs), so this needs new code.
- On timeout: mark the job `timed_out` via `DvmJobRepository.updateStatus()` and kill the worker process (soft Node-level guard  timeout + `child.kill()`, not full OS-level sandboxing; that's  a stretch goal, not required for this).
- On success: mark the job `completed`, publish a kind 6000-6999 job result event.
- On worker error/crash: mark the job `failed` with an error message.
- Tests for the dispatch flow: successful completion, timeout, worker crash, and the state  transitions in each case.

Priyanshubhartistm:
## Description

Adds dispatch for pending NIP-90 DVM job requests. `DvmOrchestratorWorker` now spawns one long-lived
worker process per configured `dvm.workers[i]` entry and multiplexes every in-flight job over that process's stdin/stdout as newline-delimited JSON:

- Polls `DvmJobRepository.findPendingJobs()` every 2s and assigns a job via the existing atomic
  `assignWorker()`.
- Sends the job's id/kind/pubkey/tags/content as one JSON line on the worker's stdin.
- Correlates the worker's `{ id, content }` reply back to the pending job by id.
- On timeout: marks the job `timed_out` and kills the worker process (any other jobs still in flight on it fail as a side effect; the worker is respawned lazily on the next poll tick).
- On success: marks the job `completed` and publishes a kind 6000-6999 result event, signed by the relay's own derived keypair (same self-signing pattern already used for invoice notifications in `payments-service.ts` `dvm.workers[]` has no per-worker signing key).
- On worker crash/unexpected exit: marks every job still in flight on that worker `failed`.

`spawnWorkerProcess()` is a new helper (in `process.ts`) for this a persistent-process counterpart to the existing one-shot `runCommand`/`runCommandWithOutput` (same spawn/error classification conventions: `not found`/`permission-denied`/`spawn-error`), framing messages both ways as newline-delimited JSON since the one-shot helpers buffer all output into a single string, which doesn't work once more than one job can be in flight against the same worker at a time.

## Related Issue

Part of #639. Closes #731.

## Motivation and Context

Job requests (#729) and job persistence (#727) exist, but nothing hands a `submitted` job off to a worker process yet. This is the last piece of Month 1's core job-routing pipeline (worker registry → persistence → ingestion → dispatch).


src/app/dvm-orchestrator-worker.ts
```diff
@@ -1,17 +1,48 @@
...
+import { IDvmJobRepository, IEventRepository } from '../@types/repositories'
+import { spawnWorkerProcess, WorkerProcessHandle, WorkerSpawnErrorReason } from '../cli/utils/process'
 import { createLogger } from '../factories/logger-factory'
+import { IRunnable } from '../@types/base'
 import { shutdownMetricsTelemetry } from '../telemetry/metrics'
 
 const logger = createLogger('dvm-orchestrator-worker')
 
+const POLL_INTERVAL_MS = 2000
+const DEFAULT_JOB_TIMEOUT_MS = 30000
```
cameri: Let's set this to 1 minute

Priyanshubhartistm: done

cameri (approved):

PR #756, opened 2026-09-04
feat(pow): add relay-load-aware adaptive PoW difficulty

### LINKED ISSUE #755: feat: adaptive PoW difficulty based on relay load (NIP-13)
## Is your feature request related to a problem? Please describe.
  nostream already has basic NIP-13 PoW verification it checks leading zero bits on event IDs and pubkeys against a static `minLeadingZeroBits` config value (`event-message-handler.ts`). The difficulty is fixed, so it's either too low (useless during a spam flood) or too high (annoying during quiet periods).

  ## Describe the solution you'd like
  - An adaptive difficulty layer on top of the existing PoW checks that scales the target difficulty based on real-time relay load when the relay is under heavy load the bar goes up, when things are quiet it comes back down
  - Track a relay load signal (events/second at minimum) and feed it into the difficulty calculation reuse the existing EWMA rate-tracking pattern (`ewma-rate-limiter.ts`) rather than building a second tracker, unless a suitable signal already exists in `admin-metrics.ts` / `telemetry/metrics.ts`
  - Reject events that don't meet the current threshold with a clear error message so clients know what's expected
  - Operators can configure floor and ceiling difficulty levels so the  adaptive range stays within reason
  - Tests covering the difficulty calculation and the rejection flow

Priyanshubhartistm:
## Description

  nostream already has basic NIP-13 PoW verification it checks leading zero bits on event IDs and pubkeys against a static `minLeadingZeroBits` config value. The difficulty is fixed, so it's either too low (useless during a spam flood) or too high (annoying during quiet periods).

  This adds an adaptive difficulty layer that scales the required PoW between a configured floor and ceiling based on the observed event rate, tracked per worker process using the same EWMA shape already used by the relay's rate limiter, kept purely in-process rather than Redis-backed.
 When enabled, the computed difficulty replaces the static `eventId`/`pubkey` `minLeadingZeroBits` checks entirely one difficulty applied uniformly to both. Disabled by default.

  `static-mirroring-worker.ts`'s own PoW check is left untouched  it evaluates events already accepted by an upstream relay, a different trust context the adaptive/load-aware framing doesn't obviously apply to.

  ## Related Issue

  Closes #755

  ## Motivation and Context

  A fixed PoW difficulty can't respond to actual relay load it's either a no-op during quiet periods or an unnecessary burden on legitimate users during a spam flood. This lets operators set a reasonable floor/ceiling and let the relay adjust automatically instead of hand-tuning one static number.


src/utils/adaptive-pow.ts
```diff
@@ -0,0 +1,34 @@
...
+  rate = 0
+  lastEventAt = 0
+}
+
+export const getCurrentDifficulty = (config: AdaptivePowSettings): number => {
+  if (config.targetEventsPerSecond <= 0 || rate <= config.targetEventsPerSecond) {
+    return config.floorBits
+  }
+
+  const ratio = rate / config.targetEventsPerSecond
```
phoenix-server: Blocking: units mismatch. `rate` is a recency-weighted event **count** (steady state at R sustained events/sec ≈ R * periodMs/1000/ln(2) — about 86.6x R at the default 60000 ms), but it is compared against `targetEventsPerSecond`. With the shipped defaults (target 50, floor 0, ceiling 24), simulated with this code: 0.5 sustained eps → difficulty 0; 1 eps → 18 bits; 2 eps → pinned at the 24-bit ceiling, while the docs promise the ceiling at 100 eps. The tests only pass because they use same-instant bursts and tiny targets.

Suggested fix: normalize to events per second before comparing, e.g. `const eps = rate / (config.periodMs / 1000 / Math.LN2)`, then compare `eps` against `targetEventsPerSecond` (and use it in `ratio`). Alternatively rename the setting to a per-EWMA-window unit and fix defaults + docs, but per-second is what CONFIGURATION.md already promises. Either way, please add a unit test with time-spaced events (fake timers, one per second) pinning floor-at-target so this can't regress.

src/utils/adaptive-pow.ts
```diff
@@ -0,0 +1,34 @@
...
+  const scaled = config.floorBits + Math.ceil((ratio - 1) * (config.ceilingBits - config.floorBits))
+
+  return Math.min(config.ceilingBits, scaled)
```
phoenix-server: Blocking: no floor clamp, and `validateSettings` only shape-checks, so `floorBits > ceilingBits` or negative values pass validation. With floor 20 / ceiling 10: difficulty 10 at rate 55, 0 at rate 150, -10 at rate 200 — and since the check is `pow < requiredBits`, a negative requirement never rejects, silently disabling PoW exactly under load. Same for a negative `floorBits` at low rates, and `periodMs: 0` makes `rate` NaN (all comparisons false → floor... or rather NaN propagates into `scaled`).

Suggested: `return Math.max(config.floorBits, Math.min(config.ceilingBits, scaled))`, plus semantic validation in `validateSettings`: `0 <= floorBits <= ceilingBits`, `periodMs > 0`, `targetEventsPerSecond > 0`.

src/handlers/event-message-handler.ts
```diff
@@ -195,17 +199,32 @@ export class EventMessageHandler implements IMessageHandler {
       return `rejected: created_at is more than ${limits.createdAt.maxNegativeDelta} seconds in the past`
     }
 
-    if (typeof limits.eventId?.minLeadingZeroBits !== 'undefined' && limits.eventId.minLeadingZeroBits > 0) {
+    if (limits.pow?.enabled) {
```
phoenix-server, line 212: This records the event before acceptance, so signature-valid events later rejected (PoW, blacklist, auth, NIP-05, dedup) all count toward the difficulty. Per-pubkey rate limits don't bound the aggregate, so pubkey-rotating spam can push everyone to ceiling with cheap unmined events, and ordinary client retries inflate the gate. If counting rejected load is deliberate (defensible as a CPU-cost proxy — signatures were already verified), please document that in CONFIGURATION.md; otherwise move the recording to after acceptance (or count only repository inserts).

src/@types/settings.ts
```diff
@@ -92,6 +92,19 @@ export interface EventRetentionLimits {
   pubkey?: EventRetentionPubkeyLimits
 }
 
+export interface AdaptivePowSettings {
+  /** Enables load-aware difficulty scaling on top of eventId/pubkey minLeadingZeroBits. Defaults to false. */
```
phoenix-server: Nit: "on top of eventId/pubkey minLeadingZeroBits" contradicts the behavior — the feature replaces those checks while enabled (your own test pins that a static 40 is ignored, and CONFIGURATION.md says "in place of"). Suggest: "replaces the eventId/pubkey minLeadingZeroBits checks while enabled."

CONFIGURATION.md
```diff
@@ -159,11 +159,16 @@ The settings below are listed in alphabetical order by name. Please keep this ta
 | limits.event.content[].maxLength            | Maximum length of `content`. Defaults to 1 MB. Disabled when set to zero. |
 | limits.event.createdAt.maxNegativeDelta     | Maximum number of seconds an event's `created_at` can be in the past. Defaults to zero. Disabled when set to zero. |
 | limits.event.createdAt.maxPositiveDelta     | Maximum number of seconds an event's `created_at` can be in the future. Defaults to 900 (15 minutes). Disabled when set to zero. |
-| limits.event.eventId.minLeadingZeroBits     | Leading zero bits required on every incoming event for proof of work. Defaults to zero. Disabled when set to zero. |
+| limits.event.eventId.minLeadingZeroBits     | Leading zero bits required on every incoming event for proof of work. Defaults to zero. Disabled when set to zero. Ignored while `limits.event.pow.enabled` is true. |
```
phoenix-server: Nit: "Ignored while limits.event.pow.enabled is true" is only true for the client path — `static-mirroring-worker.ts` still enforces the static `minLeadingZeroBits` for mirrored events (arguably correct: mirrors shouldn't demand adaptive mining). Worth a parenthetical so operators running mirrors aren't surprised.

CONFIGURATION.md
```diff
@@ -159,11 +159,16 @@ The settings below are listed in alphabetical order by name. Please keep this ta
...
+| limits.event.pow.ceilingBits                | Maximum adaptive PoW difficulty, reached at 2x `targetEventsPerSecond` and beyond. |
```
phoenix-server, line 165: Nit: the table header asks for alphabetical order, but these four `pow` rows sit between `eventId` and `kind` (kind < pow). Please move them after the `limits.event.pubkey` rows.

chappie-daemon (changes requested):
## Review: request changes

Independent two-axis review (correctness/security + conventions). The feature design is sound and well-documented (soft per-worker gate, opt-in, disabled by default, changeset + CONFIGURATION.md present), CI is green, and the tests pin real boundaries. But the core mapping does not do what its name, docs, and defaults promise, and two integration points were missed. Details below; the two blocking items are #1 and #2.

### Blocking

**1. The difficulty mapping is mis-scaled ~87x at defaults.** `recordEvent` feeds each event to `calculateEWMA` with `step=1` and no time normalization, so `rate` is a recency-weighted event **count** (steady state at R sustained events/sec is `R * periodMs/1000/ln(2)`, about 86.6 * R at the default 60000 ms half-life), but `getCurrentDifficulty` compares it directly against `targetEventsPerSecond`, a per-second rate. Simulated with the PR code and shipped defaults (target 50, floor 0, ceiling 24): 0.5 sustained events/sec → difficulty 0; 1 event/sec → 18 bits; 2 events/sec → pinned at the 24-bit ceiling. The docs promise the ceiling at 2x target = 100 events/sec. On any live relay this degenerates into a permanent 24-bit requirement (~16.7M hashes/event). The tests pass only because they use same-instant bursts and tiny targets. Suggested fix: normalize before comparing (`eps = rate / (periodMs / 1000 / Math.LN2)`), or rename the setting to a per-EWMA-window unit, fix the default and docs to match, and add a unit test with time-spaced events (fake timers, one per second) pinning floor-at-target.

**2. No floor clamp / no semantic config validation.** `getCurrentDifficulty` clamps at `ceilingBits` but never at `floorBits`, and `validateSettings` only shape-checks against default-settings.yaml. With `floorBits: 20, ceilingBits: 10`: difficulty 10 at rate 55, **0** at rate 150, **-10** at rate 200 — since the check is `pow < requiredBits`, a negative requirement never rejects, so PoW silently disables exactly when the relay is loaded. Suggested fix: `Math.max(config.floorBits, Math.min(config.ceilingBits, scaled))`, plus semantic validation (0 <= floorBits <= ceilingBits, periodMs > 0, targetEventsPerSecond > 0) in `validateSettings`.

### Major

**3. NIP-11 metadata not updated** (`src/handlers/request-handlers/root-request-handler.ts`, outside this diff): `min_pow_difficulty` is advertised from the static `minLeadingZeroBits` only, and `restricted_writes` is computed from the static bits only. With adaptive enabled and static bits unset, clients see no PoW requirement while events start getting rejected under load. Suggest advertising `floorBits` when `pow.enabled` (noting the live requirement can be higher) and counting `pow.enabled` toward `restricted_writes`.

**4. `recordAdaptivePowEvent` fires before acceptance**, so signature-valid events rejected for PoW, blacklist, auth, NIP-05, or dedup all count toward difficulty. Per-pubkey rate limits do not bound the aggregate, so an attacker rotating pubkeys can push every user to ceiling difficulty with cheap unmined spam, and ordinary client retries inflate the gate. If counting rejected load is deliberate (defensible: it is a CPU-cost proxy, signatures were already verified), please document that semantics in CONFIGURATION.md; otherwise record only accepted events.

### Minor / nits (inline below)

- CONFIGURATION.md: the four `limits.event.pow` rows break the table's own alphabetical-order rule.
- `src/@types/settings.ts`: TSDoc says "on top of" the static checks, but the feature replaces them while enabled.
- CONFIGURATION.md: "ignored while pow.enabled is true" is only true for the client path; `static-mirroring-worker.ts` still enforces static bits for mirrored events. Worth a parenthetical for mirror operators.

### Verified fine

EWMA cold-start (the `lastEventAt = 0` first-call trick), inclusive-at-target floor, `Math.ceil` rounding (defender-favorable), ceiling clamp, per-worker in-process design (deliberate and disclosed in the changeset), config plumbing and defaults, and the new unit tests otherwise pin real boundaries. The test gap that let #1/#2 through CI: no time-spaced events and no floor-over-ceiling case.

CONFIGURATION.md
Priyanshubhartistm: Done

CONFIGURATION.md
Priyanshubhartistm: Done

src/handlers/event-message-handler.ts
```diff
@@ -127,6 +131,16 @@ export class EventMessageHandler implements IMessageHandler {
       return
     }
 
+    // Recorded here, not inside canAcceptEvent's PoW branch: only events that
+    // clear every admission check (PoW, blacklist, auth, NIP-05, dedup, ...)
```
phoenix-server: Dedup is not actually cleared before this point -- the strategies only learn whether the write landed (`count ? '' : 'duplicate:'`) inside `strategy.execute`, which runs *after* the recording. So replayed/duplicate events still count toward the load signal. Either drop `dedup` from the comment, or move the recording after `strategy.execute`.

src/utils/adaptive-pow.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const getCurrentDifficulty = (config: AdaptivePowSettings): number => {
+  const eventsPerSecond = getCurrentEventsPerSecond(config.periodMs)
+
+  if (config.targetEventsPerSecond <= 0 || eventsPerSecond <= config.targetEventsPerSecond) {
+    return config.floorBits
+  }
+
+  const ratio = eventsPerSecond / config.targetEventsPerSecond
+  const scaled = config.floorBits + Math.ceil((ratio - 1) * (config.ceilingBits - config.floorBits))
```
phoenix-server, line 39: This single `requiredBits` is also applied to the pubkey check in `canAcceptEvent`, but `getPubkeyProofOfWork` cannot be mined per-event the way an event id can -- it requires generating fresh keypairs until the public key carries that prefix. With the shipped `ceilingBits: 24` that is ~16.8M keypairs (~2,500/sec with the Node stdlib on this box), and an existing identity can never comply retroactively, whatever it pays in event-id mining. Under sustained accepted load, every author whose key lacks the prefix is simply locked out. Suggest keeping the static `pubkey.minLeadingZeroBits` check while adaptive PoW is enabled, or adding a separate `pubkeyCeilingBits`. Detail in the review body.

chappie-daemon (changes requested):
## Review: request changes (re-review of e454e26)

I re-reviewed the fix commit. All four items from my previous review are resolved — verified by reading the head and running the affected suites locally on `e454e26` (74 passing across `adaptive-pow.spec.ts` / `settings-config.spec.ts` / `root-request-handler.spec.ts`, plus 192 in `event-message-handler.spec.ts`). One new blocking item, two minors, three nits.

### Verified resolved

- **#1 units.** `getCurrentEventsPerSecond` divides the EWMA count by `periodMs / 1000 / ln2`. Reproduced with your own module: sustained 1 / 5 / 25 / 50 eps read back as 1.0058 / 5.0058 / 25.0058 / 50.0058, and the ceiling is reached at exactly 2x target (20 eps vs target 10 → 24 bits). The mapping now does what the docs promise.
- **#2 clamp + semantic validation.** `Math.max(floorBits, Math.min(ceilingBits, scaled))`, plus `0 <= floorBits <= ceilingBits`, `periodMs > 0`, `targetEventsPerSecond > 0`, with the floor-over-ceiling case pinned by a test. I also confirmed a string `floorBits` cannot reach the arithmetic: `validateShape` type-checks against `default-settings.yaml` and rejects unknown keys, so the `"8" + 5 === "85"` path is closed.
- **#3 NIP-11.** `min_pow_difficulty` advertises `floorBits` when enabled, `restricted_writes` counts `pow.enabled`, tests pin all three permutations.
- **#4 recording point.** Moved past the admission checks; the rate-limited path is pinned by a test.
- **Nits.** TSDoc wording, the `static-mirroring-worker.ts` parenthetical, and the row order (`eventId` < `kind` < `pow` < `pubkey`) are all fixed.

### Blocking

**A. Applying the adaptive difficulty to the pubkey check turns the ceiling into an unsatisfiable identity requirement.**

`canAcceptEvent` uses one `requiredBits` for both axes:

```ts
const requiredBits = getAdaptivePowDifficulty(limits.pow)
if (getEventProofOfWork(event.id) < requiredBits) { ... }
if (getPubkeyProofOfWork(event.pubkey) < requiredBits) { ... }
```

`getPubkeyProofOfWork` is the leading zero bits of the pubkey itself, so it cannot be mined per-event the way an event id can — it requires generating *fresh keypairs* until the public key has that prefix. Measured on this box, `crypto.createECDH('secp256k1')` generates ~2,500 keys/sec (~2,500 is the stdlib rate; native secp256k1 libraries are several times faster), so the shipped default ceiling of 24 bits is ~16.8M keypairs — on the order of an hour of dedicated single-core keygen, and an *existing* identity can never comply retroactively, no matter how much it mines event ids.

Consequence with the shipped defaults (`floorBits: 0`, `ceilingBits: 24`, `targetEventsPerSecond: 50`): under any sustained accepted load above ~2x target, every event from an author whose key lacks 24 leading zero bits is rejected, with a `pow: pubkey difficulty 3<24`-style reason the client cannot act on. The advertised semantics — and the point of the feature — is a soft "slow down under load" gate; this is a hard write lockout for normal users until the EWMA decays. It also hands an attacker a cheap lockout: at `floorBits: 0` the events that raise the rate only need valid signatures (no mining), so a flood of ~100 signed events/sec pins the gate at ceiling and locks out honest writers for the duration of the decay — and at floor 0 the per-pubkey rate limit binds nothing, since rotating signing keys is free.

This is not covered by "applied to both eventId and pubkey checks" in the description: the reason `eventId.minLeadingZeroBits` and `pubkey.minLeadingZeroBits` are separate settings in the first place is that the two axes have wildly different costs. Collapsing them means the ceiling has to be chosen for the cheaper axis, and 24 is exactly the wrong number for the pubkey one.

Suggested fix (either is fine):

1. Keep the static `pubkey.minLeadingZeroBits` check as-is while adaptive PoW is enabled — the adaptive requirement applies to `eventId` only (smallest change, preserves the current contract for the pubkey axis); or
2. Add a separate `pubkeyCeilingBits` (defaulting to something a vanity key can plausibly meet, e.g. 8-12) and keep the pubkey requirement clamped to it.

Plus, either way: document the floor-0 amplification caveat in `CONFIGURATION.md` — with `floorBits: 0` the load signal is free to drive, so an operator who wants the gate to cost an attacker anything should set a non-zero floor.

### Minor

**B. The new comment overstates what is cleared before recording.** It says only events that clear "PoW, blacklist, auth, NIP-05, **dedup**, ..." count, but dedup happens *after* the recording point: the strategies do `eventRepository.create(event)` and only then decide `count ? '' : 'duplicate:'` (`default-event-strategy.ts:20` and its siblings). So an already-stored event replayed by a client — or a replaceable-event replay that lands as a no-op — still counts toward the load signal. Either drop `dedup` from the comment, or record after `strategy.execute` (which needs the strategy to report whether the write landed).

### Nits

**C. The rate estimator is biased high by a fixed `1000 * ln2 / (2 * periodMs)` events/sec**, so the boundaries sit slightly below the configured values. At the default 60s half-life that is +0.0058 eps: a sustained rate of exactly `targetEventsPerSecond` reads 50.0058 against a target of 50 and yields difficulty 1, not the documented floor (I reproduced this: target 10 → 9 bits at exactly 10 eps; `2x target` is effectively reached at ~1.988x). Negligible at the default target (~0.01%), ~0.6% if an operator configures a 1 eps target. Either say "approximately" in the docs, or assert the boundary with a tolerance in the test (the discrete recurrence `1/(1-e^{-lambda*dt})` is what the code actually converges to).

**D. `targetEventsPerSecond` TSDoc still says "same EWMA scale as `limits.event.rateLimits`".** After the normalization those units are real events/sec, whereas the rate limiter's `rate` setting is a raw EWMA count — the phrasing invites the two to be compared directly, which is exactly the bug that was just fixed. Worth rewording.

**E. `ceilingBits` has no upper bound.** NIP-13 difficulty saturates at 256; even 64 makes the gate unsatisfiable in practice. `0 <= floorBits <= ceilingBits` means a typo like `ceilingBits: 240` validates cleanly and permanently rejects every write. A `ceilingBits <= 256` check (or a warning) is cheap insurance.

### CI

Every check passes except `Coveralls - Integration` (coverage decreased 0.05% → 33.798%), which is why the merge state reads UNSTABLE. Not a test failure — the new source lines dilute integration coverage and no integration test covers them. The unit-coverage checks pass (69.56%, +0.2%).

src/handlers/event-message-handler.ts
Priyanshubhartistm: Done

chappie-daemon (approved):
## Review: approve (re-review of 4326c2a)

Re-reviewed the fix commit `106e398` plus the `main` merge (`4326c2a`). Blocking item A is resolved, and the branch is current with `origin/main`. Verified by reading the head and by running the suites in a clean worktree, not by trusting CI alone.

### Verified resolved

- **A. Adaptive difficulty is scoped to the event id.** `canAcceptEvent` now gates only `getEventProofOfWork(event.id)` on `limits.pow`, and `limits.pubkey.minLeadingZeroBits` remains an independent static check (`src/handlers/event-message-handler.ts:219-238`). The two new tests pin exactly the maintainer direction: with adaptive floor 9 and an 8-zero-bit pubkey but static pubkey bits unset, the event is **accepted**; with static `pubkey.minLeadingZeroBits: 16` it is rejected as `pow: pubkey difficulty 8<16`.
- **B. Recording-point comment.** Now states the actual semantics — recording happens after all admission checks, and a duplicate/no-op write still counts (dedup is decided later inside the strategy). `CONFIGURATION.md` matches.
- **C. Estimator bias.** Docs and TSDoc now say "approximately"; the boundary tests use rates safely under target and reach the ceiling via the clamp, so the discrete-recurrence overshoot no longer contradicts the docs.
- **D. Units wording.** TSDoc now reads "Event-rate threshold, in real events/sec".
- **E. `ceilingBits` bound.** `ceilingBits <= 256` validation added, with tests pinning 256 accepted and 300 rejected.
- **Merge.** `origin/main` is an ancestor of the head; the net diff against `main` is the 12 feature files with no conflict residue.

### Verification on `4326c2a`

- Affected suites: **270 passing** (`adaptive-pow`, `settings-config`, `root-request-handler`, `event-message-handler`).
- Full unit suite (excluding `test/unit/cli`, which needs a built `dist/`): **1796 passing, 0 failing**.
- `biome lint` clean on the changed files; `tsc --noEmit` reports only the pre-existing missing `@cucumber/cucumber` type-lib entry, identical on `main`.
- CI on head: everything green except `Coveralls - Integration` (−0.04% → 33.833%), which is coverage dilution from the new source lines, not a test failure — that check is what keeps the merge state at UNSTABLE.

### Nit (non-blocking)

`.changeset/adaptive-pow-pipeline.md:9` still says the scaling applies "in place of the existing static `minLeadingZeroBits` values" — plural. Only `eventId.minLeadingZeroBits` is replaced now; `pubkey.minLeadingZeroBits` is still enforced. The PR description carries the same stale sentence ("replaces the static `eventId`/`pubkey` `minLeadingZeroBits` checks entirely one difficulty applied uniformly to both"). Worth correcting both so the released changelog entry does not contradict `CONFIGURATION.md`.

Approving — the remaining item is documentation wording, not a behavior gap.

src/utils/adaptive-pow.ts
Priyanshubhartistm: done

chappie-daemon (approved):
## Re-review of `3eec7aa`: approved

The delta since my approval (`4326c2a`) is one commit, one file: `.changeset/adaptive-pow-pipeline.md` (+5 −3). `gh api compare 4326c2a...3eec7aa` confirms no source, test, or config file moved, so the verification recorded in the previous review — 270 passing across the four affected suites, 1796 passing in the full unit run, `biome` and `tsc` clean — still describes this head. CI re-ran it independently: **Unit Tests And Coverage, Integration Tests, Lint code, Lint commits, Build check, Changeset Required all pass on `3eec7aa`.**

### The nit is fixed, and correctly

The changeset now says the scaling applies "in place of the existing static `eventId.minLeadingZeroBits` value (the static `pubkey.minLeadingZeroBits` check is unaffected and still enforced independently)". That matches the code at this head — `src/handlers/event-message-handler.ts:219-238`, where the adaptive branch gates `getEventProofOfWork(event.id)` and the `pubkey.minLeadingZeroBits` check below it is unconditional — and it matches `CONFIGURATION.md:167`/`:172`. The released changelog entry no longer contradicts the documentation.

### One cosmetic item left (non-blocking)

The PR description still carries the stale sentence: *"When enabled, the computed difficulty replaces the static `eventId`/`pubkey` `minLeadingZeroBits` checks entirely one difficulty applied uniformly to both."* The description doesn't ship with the release, so this isn't a risk — but it now reads as contradicting the PR's own changeset. Edit it, or say the word and I'll correct it.

### Merge state

`MERGEABLE`, 0 commits behind `main`, approved. The `UNSTABLE` state is the `Coveralls - Integration` context alone (33.833%, −0.04% — dilution from the ~300 new source lines with no integration coverage pointed at them). Every test and lint check passes; the merge gate that remains is a coverage-percentage threshold, not a failing test.

cameri (approved):

